PR #18, opened 2023-11-21
create task generator

kaplanelad:
Addin task generator template.

Flow:
1. Create file in src/tasks
2. Add the file into mod.rs
3. Inject the task in app.rs
```rs
 fn register_tasks(tasks: &mut Tasks) {
        tasks.register(tasks::foo::Foo);
    }
  ```

```sh
$ rr generate task foo
$ rr task foo
    Finished dev [unoptimized + debuginfo] target(s) in 4.56s
     Running `target/debug/main task foo`
Task Foo generated
```

### DIFF
--- a/src/gen/templates/task.t
+++ b/src/gen/templates/task.t
@@ -1,15 +1,16 @@
 {% set file_name = name |  snake_case -%}
-{% set struct_name = file_name | pascal_case -%}
+{% set module_name = file_name | pascal_case -%}
 to: "src/tasks/{{file_name}}.rs"
 skip_glob: "src/tasks/{{file_name}}.rs"
-message: "?????????????????."
+skip_exists: true
+message: "A Task `{{module_name}}` was added successfully. Run with `cargo run task {{name}}`."
 injections:
 - into: "src/tasks/mod.rs"
   append: true
   content: "pub mod {{ file_name }};"
 - into: src/app.rs
   after: "fn register_tasks"
-  content: "        tasks.register(tasks::{{file_name}}::{{struct_name}});"
+  content: "        tasks.register(tasks::{{file_name}}::{{module_name}});"
 ---
 use std::collections::BTreeMap;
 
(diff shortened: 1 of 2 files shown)


src/gen/templates/task.t
```diff
@@ -0,0 +1,36 @@
+{% set file_name = name |  snake_case -%}
+{% set struct_name = file_name | pascal_case -%}
+to: "src/tasks/{{file_name}}.rs"
+skip_glob: "src/tasks/{{file_name}}.rs"
```
jondot, line 4: because the file name is known in advance you can use just:
```
skip_exists: true
```
instead of `skip_glob`

src/gen/templates/task.t
```diff
@@ -0,0 +1,36 @@
...
+message: "?????????????????."
```
jondot, line 5: forgot the msg

src/gen/templates/task.t
```diff
@@ -0,0 +1,36 @@
+{% set file_name = name |  snake_case -%}
```
jondot, line 1: better to name it `module_name` (file name is including `.rs`)

src/gen/templates/task.t
```diff
@@ -0,0 +1,36 @@
...
+pub struct {{struct_name}};
+#[async_trait]
+impl Task for {{struct_name}} {
+    fn task(&self) -> TaskInfo {
+        TaskInfo {
+            name: "{{name}}".to_string(),
+            detail: "Task generator".to_string(),
+        }
+    }
+    async fn run(&self, app_context: &AppContext, vars: &BTreeMap<String, String>) -> Result<()> {
```
jondot, line 32: `_app_context` and `_vars` will not give out warnings (variables not in use)

jondot (approved):

src/gen/templates/task.t
kaplanelad: changed

src/gen/templates/task.t
kaplanelad: sorry. changed

src/gen/templates/task.t
kaplanelad: good idea. fixed

PR #35, opened 2023-11-26
show list of contrllers via cli

kaplanelad:
Show list of register controllers via CLI:

```sh
[APP] controller
[GET] /
[GET] /_health
[GET] /_ping
[POST] /auth/login
[POST] /auth/register
[GET] /bench_db
[GET] /bench_hello
[GET] /echo
[GET] /user/current
```


docs-site/content/docs/the-app/controller.md
```diff
@@ -51,8 +51,8 @@ Controllers define Loco routes capabilities. In the example below, a controller
 ```rust
 use axum::routing::{get, post};
 Routes::new()
-    .add("/", get(hello))
-    .add("/echo", post(echo))
+    .add("/", get(hello), loco_rs::HttpMethod::GET)
```
jondot, line 54: better to replace axum::get with loco::get so:

```
.add("/", get(hello))
```

`get(hello)` -> MethodInfo { action: GET, handler: <handler> }`

Issue #56, opened 2023-11-30
generator: add test for mailer


jondot:
first, need to develop an approach for testing mailers

caution: there are multiple ways to do this.

(1) Rails closes a loop and captures 'deliveries', which might mean we need to create a loopback Sender, which will globally store incoming messages (we need to hook into the test lifecycle and empty it) -- to create something similar to rails.

Going uncharted territory we can:

(2) test only the `mail` function: input and output, and make sure to expose / refactor bits in mailer so that it becomes easy. in this technique nothing even gets to a Sender
(3) create ad-hoc REAL smtp receiver, with an interface to it (Rust has quite a few smtp mailbox simulators and smtp servers that can be embedded) and operate it

I would be in favor of (1) with great caution about global storage of messages, and in favor of (2) because it should be the solution with least moving parts.

(3) makes sense only if its dead simple, and contains little moving parts that are convincingly stable.

then take a clean sample test (located in tests/tasks/) and create a gen template from it.

generating a task already works, we only need to add another test template and inject it

PR #79, opened 2023-12-04
adding stub mailer for testing mailers flows

kaplanelad:
testing mailers #56


src/config.rs
```diff
@@ -169,6 +169,8 @@ pub struct EnableMiddleware {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Mailer {
     pub smtp: Option<SmtpMailer>,
+    #[serde(default)]
+    pub stub: bool,
```
jondot, line 173: notice -- this is the only thing out of the testing chain that has no "testing" cfg attribute,

jondot (approved):

src/config.rs
kaplanelad: done

Issue #132, opened 2023-12-12
generate code request_test template error


fan-tastic-z:
When executing `cargo loco generate scaffold post title:string content:text` of https://loco.rs/docs/getting-started/tour/#adding-a-crud-api, the generated code is in `tests/requests/post.rs` now code is `use blo::app::App` This should be `use {{pkg_name}}::app::App;` 
<img width="1280" alt="image" src="https://github.com/loco-rs/loco/assets/142720801/1e38cfa9-12bd-4f9c-bc5a-7a1890de7aa5">

fan-tastic-z:
Can I submit a PR to fix it?

jondot:
Yes, we'd love your help, thanks!

PR #133, opened 2023-12-12
fix: loco generate reqeust test template use pkg_name

fan-tastic-z:
fix https://github.com/loco-rs/loco/issues/132

### DIFF
--- a/src/gen/mod.rs
+++ b/src/gen/mod.rs
@@ -103,7 +103,7 @@ pub fn generate<H: Hooks>(component: Component) -> Result<()> {
             println!("{}", scaffold::generate::<H>(&rrgen, &name, &fields)?);
         }
         Component::Controller { name } => {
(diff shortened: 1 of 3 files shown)


jondot:
You probably want to add `pkg_name` to the [template context](https://github.com/loco-rs/loco/blob/master/src/gen/mod.rs#L106), like done here:

https://github.com/loco-rs/loco/blob/master/src/gen/mod.rs#L111

fan-tastic-z:
> template context

thanks,I update pr code

jondot:
@fan-tastic-z thanks, there is one additional place here: https://github.com/loco-rs/loco/blob/master/src/gen/scaffold.rs#L52

fan-tastic-z:
@jondot Thank you very much for your help. I just started learning Rust and found loco to be a great project. I hope I can learn Rust better through loco.

PR #150, opened 2023-12-14
[DRAFT] feat: try automod

jondot:
This shows how automod ergonomics feel.
I'm still undecided about getting rid of `mod.rs`, I've become attached to those as the "central common/utils files per folder"


schungx:
Will it throw off rust analyzer?

PR #157, opened 2023-12-16
fix: don't use package name in request test

DanielMSchmidt:
This removes an error where app_name is not correctly set.

In my app the name was set to `blo::app::App;` instead of `mealplanning::app::App;`, but `crate::app::App;` solves this without relying on a hard-coded name.


kaplanelad:
Hi @DanielMSchmidt, based on my observation, it appears that including the injection file in the `tests/` folder requires specifying the crate package name. The crate lib name is mandatory in what has under `tests` folder and not recognize `crate::`

PR #204, opened 2023-12-21
minor changes

ohaddahan:


scripts/setup_db.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
jondot, line 1: this looks good
can you convert `scripts/setup_db.sh` to `xtask`?

it basically means something like:

```
$ cargo xtask setup_db
```

Also see if you can print a notice that "A postgres database has been started in the background"

In addition moving to `xtask` means you write pure rust, so maybe you can do something like

```
$ cargo xtask setup_db --user loco --pass loco --db loco_app
```

and if no flags given, use default values

jondot: xtask code starts here: https://github.com/loco-rs/loco/blob/master/xtask/src/bin/main.rs

DEVELOPMENT.md
```diff
@@ -5,65 +5,76 @@ Errors are done with `thiserror`. We adopt a minimalistic approach to errors.
...
 ### Context
 
-When you know a user might need context, resort to manually shaping the error with extra information. First, define the error:
+When you know a user might need context, resort to manually shaping the error with extra information. First, define the
+error:
 
 ```rust
-    #[error("cannot parse `{1}`: {0}")]
-    YAMLFile(#[source] serde_yaml::Error, String),
+    # [error("cannot parse `{1}`: {0}")]
```
jondot, line 29: something strange with formatting here?
we have a space after `#`, `&`, etc.

scripts/setup_db.sh
ohaddahan: BTW usage of default PG port isn't recommended in my point of view.
I have PG installed and I always have issues with it since the docker and my local install are supposed to use the same port.

Using non-default ensures it's a smoother ride for any user, regardless if they have PG locally or not.

DEVELOPMENT.md
```diff
@@ -5,65 +5,76 @@ Errors are done with `thiserror`. We adopt a minimalistic approach to errors.
 * We try to have _one error kind_ for the entirety of Loco.
 * Errors that cannot be handled, are _informative_ and so can be opaque (we don't offer deep matching on those)
 * Errors that can be handled and reasoned upon should be able to be matched and extract good knowledge from
-* To users, error should _not be cryptic_, and should indicate how to fix issues as much as possible, or point to the issue precisely
-
+* To users, error should _not be cryptic_, and should indicate how to fix issues as much as possible, or point to the
+  issue precisely
```
kaplanelad, line 9: Why moved into a new line?

DEVELOPMENT.md
```diff
@@ -5,65 +5,76 @@ Errors are done with `thiserror`. We adopt a minimalistic approach to errors.
...
-In this case, we duplicate the YAML error type, leave one of those for auto conversions with `from`, where we don't have a file, and create a new specialized error type with the file information: `YAMLFile`.
+In this case, we duplicate the YAML error type, leave one of those for auto conversions with `from`, where we don't have
+a file, and create a new specialized error type with the file information: `YAMLFile`.
 
 ## Publishing a new version
 
 **Test your changes**
 
 * [ ] run `cargo test` on the root to test Loco itself
+* [ ] ```$ docker run -d -p 5432:5432 -e POSTGRES_USER=loco -e POSTGRES_DB=loco_app -e POSTGRES_PASSWORD="loco" postgres:15.3-alpine```
```
kaplanelad, line 51: what are you trying to say here?

scripts/setup_db.sh
ohaddahan: Removed, seems like a local issue in my env.

DEVELOPMENT.md
ohaddahan: Added more text, to to notify need the `docker` up, it's in the main `README.md` but not here, so adding for completeness.

DEVELOPMENT.md
ohaddahan: Fixed.

DEVELOPMENT.md
ohaddahan: Auto formatter going wild, reverted.

DEVELOPMENT.md
kaplanelad: While it's beneficial to enhance our text content,I believe we should aim to avoid command duplicates for improved maintainability. Consider the PR #216 as an example, we have new feature that  dynamically generate the database name based on the application name and environment in the starter. 

To optimize this, how about incorporating a concise instruction like:

"Ensure you have the necessary local resources, such as db/redis, by executing the command `cargo loco doctor`. In case you don't have them, refer to the relevant documentation section for guidance."

WDYT?

ohaddahan: Updated.

DEVELOPMENT.md
```diff
@@ -45,6 +45,7 @@ In this case, we duplicate the YAML error type, leave one of those for auto conv
 **Test your changes**
 
 * [ ] run `cargo test` on the root to test Loco itself
+* [ ] Ensure you have the necessary local resources, such as `DB`/`Redis`, by executing the command `cargo loco doctor`. In case you don't have them, refer to the relevant documentation section for guidance.
```
kaplanelad, line 48: Thank. last changes:

1. Can you bump this bullet to be first? we should validate the resources before running the test.
2. add --environment test to `cargo loco doctor`. we need to check the connectivity via testing and not the default (which is development) 
should be `cargo loco doctor --environment test`

ohaddahan: Changed.

kaplanelad (approved):
Thanks @ohaddahan

PR #205, opened 2023-12-21
Compression layer

ohaddahan:


src/controller/app_routes.rs
```diff
@@ -202,6 +203,12 @@ impl AppRoutes {
             }
         }
 
+        if let Some(cors) = &ctx.config.server.middlewares.compression {
+            if cors.enable {
```
jondot, line 207: note that this is still `cors`, should be something like `compression`

examples/demo/config/development.yaml
```diff
@@ -109,7 +112,7 @@ database:
 # Redis Configuration
 redis:
   # Redis connection URI
-  uri: {{ get_env(name="REDIS_URL", default="redis://127.0.0.1") }}
+  uri: { { get_env(name="REDIS_URL", default="redis://127.0.0.1") } }
```
jondot, line 115: note auto formatting breaks the templating format, it should be `{{`. try "save without formatting" for YAML

Cargo.toml
```diff
@@ -32,10 +32,10 @@ colored = "2"
 
 
 sea-orm = { version = "0.12.4", features = [
-  "sqlx-postgres",        # `DATABASE_DRIVER` feature
-  "sqlx-sqlite",
-  "runtime-tokio-rustls",
-  "macros",
+    "sqlx-postgres", # `DATABASE_DRIVER` feature
```
jondot, line 35: what kind of formatter do you use for `toml`?

src/config.rs
```diff
@@ -142,6 +142,8 @@ pub struct Workers {
 /// Represents the middleware configuration structure.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Middlewares {
+    /// Middleware that enable compression for the response.
+    pub compression: Option<EnableMiddleware>,
```
jondot, line 146: looks like the correct location. instead of `EnableMiddleware` maybe you want `CompressionMiddleware` ?

Cargo.toml
ohaddahan: `RustRover` default, think it supposed to use `cargo fmt` not sure if I touched that setting.

src/config.rs
ohaddahan: Isn't `EnableMiddleware` the generic `boolean` type?
Doesn't matter to me, whichever you think is best.

src/controller/app_routes.rs
ohaddahan: ops copied over :) 
Fixed.

src/config.rs
jondot: yes, my bad let me rephrase

* if you have 0 configuration parameters - yes use EnableMiddleware which only controls switching on/off the middleware
* if you have additional specific configuration for the middleware, use your own type (and add a similarly named boolean for enable)

I was almost sure a compression middleware has some custom configuration but i may be mistaken -- its your call to decide

ohaddahan: Technically the compression layer can be dissected into supported formats but I don't think we care that much and just enabling it as a whole is simple and good.

examples/demo/config/development.yaml
ohaddahan: Fixed it, seems like `JetBrains` treats `{{` as two `{` and not as a special template, had to manually fix it (no formatting option support it :( )

src/controller/app_routes.rs
```diff
@@ -178,6 +179,12 @@ impl AppRoutes {
             }
         }
 
+        if let Some(compression) = &ctx.config.server.middlewares.compression {
```
nils-werner, line 182: Not related/limited to this PR, but does `AppRoutes` really need to know about all available middlewares? What about future custom middlewares that aren't part of loco?

jondot: That's a really great question. Actually we really need more of those questions for deliberating around opinions/decisions!

The idea is that we have a set of "blessed" middleware that are strongly tied, and now we can say -- which are the ones that are really in say 80% of projects (just 80/20) rule to include, also assuming they "cost" us very little (in binary size, dependencies, and maintenance) but provide the user maximum value.

We want to keep the list very small as you say. We also assume compression is such a middleware.
What are your thoughts specifically? now that you have the context, do you think adding compression as a "first-class" is too heavy?

jondot (approved):

Issue #82, opened 2023-12-05
loco-cli should auto-install sea-orm-cli if missing


jondot:
Probably wise to also ask before installing it, and explain why it is needed.

AbhilashGade:
Can I pick this issue?

kaplanelad:
@jondot maybe we can fix the seaorm cli crate and make the generator entry to use via the crate instead of the cli installation?

what do you think?

kaplanelad:
The problem is that the sea-orm is needed only on specific templates (saas).
We could do something smarter and check in the lib that:
Check if the user uses the DB feature and the migration commands in the configuration are enabled (migration, truncate, reset..) and then check if seaorm-cli is installed. and if not, stop the service and prompt a pretty message.

What do you think?

jondot:
yes sounds good!

kaplanelad:
@AbhilashGade you can feel free and take it. 
you can implement it in the lib instead of the CLI. 
see that the validation should be only when `with-db` feature is enable

kaplanelad:
> @kaplanelad yeah on it

just make sure that you not installing automatically sea-orm. i think it is better to return an pretty error with explanation

AbhilashGade:
Okay how do I test this

kaplanelad:
@AbhilashGade you can use (check_seaorm_cli)[https://github.com/loco-rs/loco/blob/master/src/doctor.rs#L132] which doing the checks that you needs

AbhilashGade:
So do I do something like this 

`fn check_sea_orm() {
    println!(" SeaORM is not installed. SeaORM is a powerful ORM for Rust that provides convenient database abstractions. It is required for this feature. Would you like to install it? (y/n)");

    let mut input = String::new();
    std::io::stdin().read_line(&mut input).unwrap();
    let input = input.trim().to_lowercase();

    if input == "y" || input == "yes" {
        println!("Installing SeaORM... ");
        std::process::Command::new("cargo")
            .arg("install")
            .arg("sea-orm")
            .output()
            .expect("Failed to install SeaORM");
    } else {
        println!("SeaORM is required for this feature. Please install it to continue.");
    }
}
`

kaplanelad:
> > @kaplanelad yeah on it
> 
> just make sure that you not installing automatically sea-orm. i think it is better to return an pretty error with explanation

Did you see my previous message? I believe there's no need to install the sea-orm-cli. Handling the installation process might be unnecessary and open to potential problems. 
[...]

PR #210, opened 2023-12-21
Use SQLite by default

nils-werner:
Currently trying out the starter examples is quite an intensive task, due to the requirement for Postgres. In production there is definitely the need for something as scalable as Postgres, but I don't think that's the case for the starters. To lower the initial hurdle, both Rails and Django opt for SQLite by default, which is enough to hit the ground running and start playing with the frameworks.

I'd recommend loco do the same, and this PR implements that. It replaces Postgres with SQLite in the configs, and mentions the eventual need for Postgres in the readme and docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,8 @@ myapp
 ```
 
 <div class="infobox">
-To configure a database , please run a local postgres database with <code>loco:loco</code> and a db named <code>loco_app</code>.
+<code>loco new</code> uses SQLite by default. In production you should replace this with a more scalable database system, like Postgres.
 </div>
-
-You can use Docker to run a Postgres instance:
-
-```
-$ docker run -d -p 5432:5432 -e POSTGRES_USER=loco -e POSTGRES_DB=loco_app -e POSTGRES_PASSWORD="loco" postgres:15.3-alpine
-```
 
 Now `cd` into your `myapp` and start your app:
 
--- a/docs-site/content/docs/getting-started/guide.md
+++ b/docs-site/content/docs/getting-started/guide.md
@@ -52,8 +52,7 @@ Make sure you also have locally installed or running (via Docker or otherwise):
 - Redis
 
 <div class="infobox">
-To configure a database , please run a local postgres database with <code>loco:loco</code> and a db named <code>loco_app</code>: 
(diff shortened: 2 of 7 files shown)


kaplanelad:
Hey @nils-werner, 

Thanks for that. 
we started with SQLite but we have some problems with dateTime in sea-orm. so we decided to back to postgress as a default. 

try to create a new loco starter (saas/api) and try to run `cargo loco db entities` and you can see that dateTime type not converted right.

nils-werner:
Actually, `db entities` doesn't work for me at all:

```
2023-12-21T15:51:47.539516Z  WARN loco_rs::boot: entities:
Error: No such file or directory (os error 2)

Location:
    /home/nils/.cargo/registry/src/index.crates.io-6f17d22bba15001f/loco-rs-0.1.8/src/cli.rs:287:13
```

kaplanelad:
You need to install `sea-orm-cli` - cargo install sea-orm-cli
you can see it in our tour https://loco.rs/docs/getting-started/tour/

nils-werner:
Ah, duct not saying "command X not found" but "file not found" is really confusing here 😄

jondot:
The issue we had with SQLite being the default was that SeaORM "loses information" meaning the migration specifies DateTime, but SQLite will not preserve the type information. Then, when sync'ing back from the schema, we get DateTime columns serialized back as String into the entities.

There is an issue open in SeaORM for this. So in the meanwhile we decided to 

- give a command to run pg on docker
- create a vscode devcontainer
- generally recommend setting up pg locally

Regarding dealing with the SQLite issue head-on, I think our effort was like 4/10, we tried a bit but can't say we stopped the world and said: no matter what we need to fix either SeaORM or build a hacky workaround.
So there might be some solutions we did not explore for this yet.

nils-werner:
> There is an issue open in SeaORM for this.

[This one](https://github.com/SeaQL/sea-orm/issues/924)? Just want to make sure all the threads are tied together 😊

jondot:
Yes! :)

kaplanelad:
@jondot regarding the better error message we have this issue: #82.

jondot:
Yep, but then on its own, `duct` produces cryptic errors :)

kaplanelad:
@nils-werner, I appreciate your contribution to this pull request. I'm closing it for now. I would like to continue the discussion to explore potential solutions. We can either open a new issue or discuss it further in our Discord channel.

Thank you for your input. I believe your can us with other issues that we have or new ideas for improvment

Issue #120, opened 2023-12-11
refactor tracing subscriber init (logger::init) in test mode


jondot:
In test mode, it is better to externalize logger init (and global subscriber set up), and initialize once, before all tests start for the testable binary at hand.

In return, we need to do a conditional cfg(test) for logger init, inside `boot`, so it will not happen at all when in test mode

PR #213, opened 2023-12-21
wip: disable init logger in testing mode

kaplanelad:
Following issue: #120 we need to disable the logger when loco running in test mode. to avoid this error:
```
a global default trace dispatcher has already been set
```

### DIFF
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -10,6 +10,10 @@ use tracing::{trace, warn};
 
 #[cfg(feature = "with-db")]
 use crate::db;
+
+#[cfg(not(feature = "testing"))]
+use crate::logger;
+
 use crate::{
     app::{AppContext, Hooks},
     banner::print_banner,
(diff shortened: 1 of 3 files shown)


src/boot.rs
```diff
@@ -181,6 +184,8 @@ pub async fn create_context<H: Hooks>(environment: &str) -> Result<AppContext> {
     let environment = Environment::from_str(environment)
         .unwrap_or_else(|_| Environment::Any(environment.to_string()));
     let config = environment.load()?;
+
+    #[cfg(not(feature = "testing"))]
```
jondot, line 188: I think we can just set "testing = false" in config for test by default,
but then, we need to find the right point to do init in tests only once 

The challenge is finding the right spot to do init only once for the lifetype of a test -- maybe lazy_static

kaplanelad: Yes i agree, i have more problems when did it like `dead_code` in clippy ignoring. 

this is why i move it to WIP

Issue #143, opened 2023-12-13
Scaffold generates broken code in some failure cases


BWStearns:
**Description**

`cargo loco generate scaffold logs file_name:string hash:string created_at:ts`

This will fail to generate the `model` and `_entities` files because `created_at` is already included (perhaps this could be documented), but the command will still generate controller and view which will be broken. There is some warnings displayed but they're not super intuitive without having already seen this bug.

**To Reproduce**

`cargo loco generate scaffold logs file_name:string hash:string created_at:ts`

**Expected Behavior**

The command either succeeds, or displays an error and doesn't make changes.

If a default column is specified by the user it could replace the column or it could just error (the latter might be more straightforward). I think either behavior would be fine. 

**Environment:**

MBP

**Additional Context**

I haven't had time to dig into the code generation too much but afaict scaffold only generates whole new files with the exception of `mod.rs` files so maybe it could keep the original `mod.rs` contents in memory and there could be an unwind action on failures that removes created files and reverts the `mod.rs`s

jondot:
Indeed a special case! thanks for catching this!
I can think of two solutions:

1. check a field existence before adding (using the various injection tools we have)
2. keep a list of `created_at`, etc. that we don't allow from the CLI when it's generating models because we already attach these by default (I'm in favor of this one, all generators give you created/updated for free)

would you like to contribute a solution?

BWStearns:
I’d be happy to but it might be a minute. I might be able to get a PR in
next week or this weekend if I get lucky.

kaplanelad:
@BWStearns be aware that i already fix it

BWStearns:
Cool!

BWStearns:
@kaplanelad fixed it

kaplanelad:
@BWStearns, no worries! Just wanted to mention that this pull request hasn't been merged yet. I'm waiting for the CI to pass.

PR #214, opened 2023-12-21
fix date generators

kaplanelad:
Addressing issue #143, I've resolved the generator model and scaffold problem when specifying "ts/ts!"

By default, we generate created_at and updated_at fields. However, users might not be aware of this feature when using loco, and when manually adding these fields during table generation, conflicts arise in the files. To tackle this, I introduced a constant named`IGNORE_FIELDS` that acts as a blacklist for fields we should ignore.

In the controller, when a user includes ts or ts, the type of those fields is DateTime which we are not importing in the injection. i have added it into our preloader


slyons:
You might also want to print a message or warn the user if they specify either of those columns in `cargo loco generate model`, speaking as a user who just ran into this.

kaplanelad:
@slyons can you give me an example of what are you expected to see in the logs?

jondot:
@kaplanelad maybe here: https://github.com/loco-rs/loco/pull/214/files#diff-16d1f5545eaa1d32fdd08f0fa75e88e26c3a54d5e264d71fe7eeed68880a0b42R51 

say something like

```
warn!(field=field, "note that a redundant field was specified, it is already generated automatically")
```

kaplanelad:
Done.
thanks @slyons

src/cli.rs
```diff
@@ -299,6 +299,7 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
             let environment = Environment::from_str(&environment)
                 .unwrap_or_else(|_| Environment::Any(environment.to_string()))
                 .load()?;
+            logger::init::<H>(&environment.logger);
```
jondot, line 302: is there a chance we need to rethink logger init? 
like move this to the top right after main(), and put it only there (remove all init from context creation, etc)

this might be also the same pain point with logs in tests?

kaplanelad: i think we need to init the logger in the cli.rs file

Issue #115, opened 2023-12-10
token based API auth


jondot:
A `User` should have an `api_token` field, and we should be able to configure the authentication module to use it with `Bearer` authentication. Use `APIToken` extractor.

PR #217, opened 2023-12-22
Add token based api auth

kaplanelad:
token based API auth #115


examples/demo/src/controllers/middleware/api_key.rs
```diff
@@ -0,0 +1,42 @@
...
+    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Error> {
+        // Extract API key from the request header.
+        let api_key = auth::extract_token_from_header(&parts.headers)
+            .map_err(|e| Error::Unauthorized(e.to_string()))?;
+
+        // Convert the state reference to the application context.
+        let state: AppContext = AppContext::from_ref(state);
+
+        // Retrieve user information based on the API key from the database.
+        let user = users::Model::find_by_api_key(&state.db, &api_key).await?;
```
jondot, line 38: this will return "not found", but in the context of api auth, i feel it should return "not authorized", so just translate that error.

NOTE: its better to translate that error right here instead of have the model returning "not authorized" because auth/authn is the concern of this extractor

examples/demo/src/controllers/middleware/api_key.rs
```diff
@@ -0,0 +1,42 @@
...
+    http::request::Parts,
+};
+use loco_rs::{app::AppContext, controller::middleware::auth, errors::Error};
+use serde::{Deserialize, Serialize};
+
+use crate::models::_entities::users;
+
+#[derive(Debug, Deserialize, Serialize)]
+// Represents the data structure for the API token.
+pub struct ApiToken {
```
jondot, line 15: I am wondering,
how possible is it to say:

```
pub struct ApiToken<T>
 ... everything depends on T
 ... includng the extractor 
```

and we know that T should eventually be users::Model,
maybe this T will be a `T: Trait` where `T: FetchByApiKey`,

and the `User impl` that FetchByApiKey.

Now this middleware becomes shareable by everyone

examples/demo/src/controllers/middleware/api_key.rs
kaplanelad: Yes, my bad. missed it

examples/demo/src/controllers/middleware/api_key.rs
kaplanelad: The API token is generated within the user app (Loco framework is not familiar with that). In this scenario, I think the <T> parameter is meaningless, as the user has complete control over it.

About the FetchByJWT i totally agree. maybe we can consider moving the claims to be a <T>

jondot: Notice what I did in the latest commit. I had a feeling if we move things around we might get something, I think it worked:

`Authenticable` trait lets us encapsulate all the logic, and then -- just refer to an extractor. Use `ApiKey<T>` and `ApiKey<users::Model>` to give Rust all the type information it needs to resolve.

On this "Authenticable" concept we can also encode the `JWT` behavior, and "capture" a user just like in api key auth.
There is a good chance that instead of having _claims_ we can have the actual user model if we use the same trick.

kaplanelad:
Looks good @jondot, I reverted the middleware that I created. but keep the migration changes. wen releasing a new version I'll add the controller to the starters

Let me know if you want more changes from my side

Issue #203, opened 2023-12-21
Deployment: add `nginx`


jondot:
Look here: https://github.com/clowzed/sero/blob/master/nginx-templates/default.conf.template

`sero` is very similar in stack: [Axum, SeaORM, Postgres](https://github.com/clowzed/sero/issues/18). They use nginx for deployment, which is something we could use as well.

The deployment should create

* nginx conf file file (look at what `sero` does)
* any local configuration tweaks that we need in order to work behind an nginx proxy (edit YAML for production, change it, write it back)
* try to do it all with our generator + code injections framework

fivehanz:
+1

arferreira:
Hey guys can I get this issue?

kaplanelad:
Hey @arferreira, sure, go for it.

PR #244, opened 2023-12-27
feat: nginx deployment

arferreira:
Implement and add to our CLI the generation of the nginx conf file for deployment as mentioned in the issue: [(https://github.com/loco-rs/loco/issues/203)]

@jondot I had some questions about whether for now we can start with the simplest implementation and wait for demand, or whether we could think about SSL and TLS. Regarding the current implementation, it is very simple looking directly at the configurations made.

### DIFF
--- a/docs-site/content/docs/getting-started/deployment.md
+++ b/docs-site/content/docs/getting-started/deployment.md
@@ -94,6 +94,8 @@ cargo loco generate deployment
 ? ❯ Choose your deployment ›
 ❯ Docker
 ❯ Shuttle
+❯ Nginx
+
 ..
 ✔ ❯ Choose your deployment · Docker
 skipped (exists): "dockerfile"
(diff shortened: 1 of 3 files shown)


jondot:
Looks great! I think simplicity is a fantastic thing :)

What do you think about looking a the docker-compose https://github.com/clowzed/sero/blob/master/docker-compose.yml and trying to customize this one to our needs? then drop the generated docker-compose file in `ngnix/` as well. This will be a great "dont make me think" feature that will streamline deploying under nginx.

Only add it if its not too complicated, if you find that it involved some jumping through hoops, we can leave it to later :)

kaplanelad:
You can also look at our docker-compose in the starters template: 
https://github.com/loco-rs/loco/blob/master/starters/saas/.devcontainer/docker-compose.yml

thomas9911:
Not to be annoying, I know that nginx is the default for this problem. But maybe it would be nice to add traefik, it integrates nicely with docker-compose (adding a couple of labels) and you can get lets encrypt certificates set up quite easily.

tjyang:
Hi
To learn more about this PR, I did a manual patching in my forked repository. 
I found there is file missed. Please review and confirm.

- deployment_nginx_config.t is missed in templates dir
image
- content of deployment_nginx_config.t
```
to: "Nginx.toml"
skip_exists: true
message: "Nginx.toml file created successfully"
---
name = "{{pkg_name}}"
```
- testing output
image

jondot:
going to move forward with this one

PR #252, opened 2023-12-27
fix: update register api log

fan-tastic-z:
saas and rest-api auth register fail , use `tracing::error!`

### DIFF
--- a/starters/rest-api/src/controllers/auth.rs
+++ b/starters/rest-api/src/controllers/auth.rs
@@ -36,7 +36,7 @@ async fn register(
     let user = match res {
         Ok(user) => user,
         Err(err) => {
-            tracing::info!(
+            tracing::error!(
                 message = err.to_string(),
                 user_email = &params.email,
                 "could not register user",
--- a/starters/saas/src/controllers/auth.rs
+++ b/starters/saas/src/controllers/auth.rs
@@ -36,7 +36,7 @@ async fn register(
     let user = match res {
         Ok(user) => user,
         Err(err) => {
-            tracing::info!(
+            tracing::error!(
                 message = err.to_string(),
                 user_email = &params.email,
                 "could not register user",


kaplanelad:
@fan-tastic-z This is not something you want to be tagged as an error.
Let's say that this error shipped into your logging system. and you have users that log in with the wrong user/pass.

which action do you need to take?

fan-tastic-z:
> @fan-tastic-z, tagged as an error in this situation, does not look like something we want to see as an error. Imagine this error entering your logging system and users attempting to log in with incorrect credentials.
> 
> What course of action would you be going to do?

@kaplanelad When I use loco to develop, I call the interface for registered users. When the registration fails, the interface returns no error message, but the log shows info. Is it better to use error? Easier to troubleshoot where the problem is

fan-tastic-z:
> @fan-tastic-z, tagged as an error in this situation, does not look like something we want to see as an error. Imagine this error entering your logging system and users attempting to log in with incorrect credentials.
> 
> What course of action would you be going to do?

image

Regardless of whether the registration is successful or failed, the return of the interface is like this

kaplanelad:
@fan-tastic-z, the resignation function currently invokes `create_with_password`, which, in turn, returns an error also if the user already exists. 
in this case, this is not a real error that you have something todo (and you don't want to return an error for a security reason). 

The code is present in the starter templates, allowing users the flexibility to modify and handle according to their preferences. Currently, the login is set to log information as per the design.

fan-tastic-z:
> @fan-tastic-z, the resignation function currently invokes `create_with_password`, which, in turn, returns an error also if the user already exists. in this case, this is not a real error that you have something todo (and you don't want to return an error for a security reason).
> 
> The code is present in the starter templates, allowing users the flexibility to modify and handle according to their preferences. Currently, the login is set to log information as per the design.

@kaplanelad Got it, thank you for your answer

PR #257, opened 2023-12-27
Move git deps to optional in CLI

Virviil:
### Changes

- Removed mandatory dependency on `git2` in the CLI, since 99% of programmers already have Git installed as CLI.
- Added optional dependency on `git2` and `gix` 
- Added the ability to choose an alternative Git via the `--feature` flag.
    - `git2`: binds to libgit.
    - `gix`: fully native Git written in Rust.

### Explanations

This change improves the installation process by removing the mandatory dependency on `git2` and giving users the option to choose between two Git alternatives. 
This can be crucial because of some issues with git2 (https://github.com/rust-lang/git2-rs/issues/672).
Now, if the user already has Git installed, the installation process will proceed more smoothly.

### Additional Changes

- Excluded `Cargo.lock` from the repository as it is not used during installation via `cargo install`.


kaplanelad:
hey, can you explain more how it improve the installation?

the default clone is based in the local git and the fallback is automatically happen to the lib.

Virviil:
@kaplanelad It improves the installation of a CLI itself.

`cargo install loco-cli` is failing for me (hope that not only for me) because of https://github.com/rust-lang/git2-rs/issues/672. git2 is mandatory, while being a fallback (as you say).

Fallback is very hardly used, because developers using loco for sure have `git` installed.

So, this change move this dep into optional.

Virviil:
@kaplanelad @jondot any updates, suggestions?

kaplanelad (approved):

PR #265, opened 2023-12-28
Add verbosity as argument for CLI

Virviil:
New CLI arg was added - `-v` | `--verbose`, that is more-less standard for CLIs, that allows to trace problems running CLI.
Before the PR one can do it just by setting `RUST_LOG` env, which is poorly documented.

### DIFF
--- a/loco-cli/src/bin/main.rs
+++ b/loco-cli/src/bin/main.rs
@@ -1,12 +1,17 @@
 use clap::{Parser, Subcommand};
 use loco_cli::{generate, git, prompt, CmdExit};
 use std::path::PathBuf;
+use tracing::level_filters::LevelFilter;
 
 use tracing_subscriber::EnvFilter;
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
 #[command(propagate_version = true)]
 struct Cli {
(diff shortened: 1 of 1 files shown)


loco-cli/src/bin/main.rs
```diff
@@ -24,8 +29,21 @@ enum Commands {
 fn main() -> eyre::Result<()> {
     let cli = Cli::parse();
 
+    println!("{:?}", &cli.verbose);
```
jondot, line 32: you probably want to remove this right?

Virviil: Y, sure. One more question - will this change break a test? Probably I should fix it also

kaplanelad: @Virviil, you are correct. See [here](https://github.com/loco-rs/loco/blob/master/loco-cli/tests/cli_tests.rs#L3)  it is commented out. 
Do you mind to fix it?

Virviil: Fixed

Virviil: @kaplanelad I'm not sure that i'll take this task, but for sure not as a part of this PR. So let's just proceed

kaplanelad (approved):

Issue #282, opened 2023-12-31
Unable to log in in Production


slyons:
**Description**

I've got my first simple app in Loco deployed to a hosting service (Fly.io). All of my unit tests pass, all my local testing works great. As soon as I deploy the Docker image the Login function stops working. I can register my users just fine, but the password hashing fails. Is there some difference  between the development config and production config that could lead to the hashing failing? I've validated that the usernames and passwords are not being modified on the way in, and I can see that it's succeeding on finding the user.

```
2023-12-31T05:18:24Z app[e2865756b4e218] sea [info]2023-12-31T05:18:24.080779Z  INFO http-request: sqlx::query: summary="SELECT \"users\".\"created_at\", \"users\".\"updated_at\", \"users\".\"id\", …" db.statement="\n\nSELECT\n  \"users\".\"created_at\",\n  \"users\".\"updated_at\",\n  \"users\".\"id\",\n  \"users\".\"pid\",\n  \"users\".\"email\",\n  \"users\".\"password\",\n  \"users\".\"api_key\",\n  \"users\".\"name\",\n  \"users\".\"reset_token\",\n  \"users\".\"reset_sent_at\",\n  \"users\".\"email_verification_token\",\n  \"users\".\"email_verification_sent_at\",\n  \"users\".\"email_verified_at\"\nFROM\n  \"users\"\nWHERE\n  \"users\".\"email\" = $1\nLIMIT\n  $2\n" rows_affected=0 rows_returned=1 elapsed=1.547908ms http.method=POST http.uri=/api/auth/login http.version=HTTP/1.1 http.user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36 environment=production request_id=54c0c6df-0753-4316-bd0d-d8ea2a0cdda2
2023-12-31T05:18:24Z app[e2865756b4e218] sea [info]2023-12-31T05:18:24.103515Z  WARN http-request: loco_rs::controller: err="unauthorized!" http.method=POST http.uri=/api/auth/login http.version=HTTP/1.1 http.user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36 environment=production request_id=54c0c6df-0753-4316-bd0d-d8ea2a0cdda2
```

```rust
[...]

PR #290, opened 2024-01-01
Adding a new command to the CLI to generate JWT secret keys

slyons:
After my own mistake with this while releasing to production I've decided that the CLI needs a way to generate new JWT secrets

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -156,6 +156,8 @@ enum ComponentArg {
     },
     /// Generate a deployment infrastructure
     Deployment {},
+    /// Generate a new JWT secret
+    JwtSecret {}
 }
 
 impl From<ComponentArg> for Component {
@@ -172,6 +174,7 @@ impl From<ComponentArg> for Component {
             ComponentArg::Worker { name } => Self::Worker { name },
             ComponentArg::Mailer { name } => Self::Mailer { name },
             ComponentArg::Deployment {} => Self::Deployment {},
+            ComponentArg::JwtSecret {} => Self::JwtSecret {  }
         }
     }
 }
--- a/src/gen/mod.rs
+++ b/src/gen/mod.rs
@@ -1,6 +1,7 @@
 use chrono::Utc;
 use rrgen::{GenResult, RRgen};
 use serde_json::json;
+use rand::{distributions::Alphanumeric, thread_rng, Rng};
 
 #[cfg(feature = "with-db")]
 mod model;
@@ -99,6 +100,7 @@ pub enum Component {
         name: String,
     },
     Deployment {},
+    JwtSecret {}
 }
 
 pub fn generate<H: Hooks>(component: Component, config: &Config) -> Result<()> {
@@ -177,6 +179,14 @@ pub fn generate<H: Hooks>(component: Component, config: &Config) -> Result<()> {
                     rrgen.generate(DEPLOYMENT_SHUTTLE_CONFIG_T, &vars)?;
                 }
             }
+        }
+        Component::JwtSecret {} => {  
+            let new_secret: String = thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(20)
+                .map(char::from)
+                .collect();
+            println!("New JWT secret: {}", new_secret);
         }
     }
     Ok(())


slyons:
Addresses #282

slyons:
Let's not merge this one yet, the problem isn't solved

kaplanelad:
Get @slyons, I don't think we should add a cli command for generating a JWT token. 
Instead, let's streamline it by incorporating the functionality under cargo generate deployment. We can enhance it further by introducing a new option, 'production YAML config.' This way, when users trigger the deployment generation, it automatically includes the generation of a configuration with a randomly generated JWT, similar to the approach we have in [loco-cli](https://github.com/loco-rs/loco/blob/master/loco-cli/src/generate.rs#L74)."

jondot:
I went through Rails docs,
It seems that they offer a generic `secret` task, where it generates a general purpose secret:

https://radavis.github.io/ruby/generate-secret-keys/

>desc "Generate a cryptographically secure secret key (this is typically used to generate a secret for cookie sessions)."

Notice - generating a secret should be clean, which means no artifacts such as "A secret was generated: <secret>" so that this can be used as a Unix tool (you can execute it anywhere you like and use the result by reading the process output). There is also no specific "JWT secret" or "Cookie secret" -- I guess a secret is a secret and they're right. We need to pick a good character distribution and a standard key length (we can use what JWT requires as a guideline for key length).

In that thinking -- generating a secret is both a dev and production general purpose _tool_.

How can we do this?

## (1) As a Task

So, rather than being part of Loco, this can be a `Task` that we provide built-in in every starter app.

This is not 1:1 to Rails, because Rails provides built-in tasks (we dont have that mechanism yet), but it is significantly similar.

```
$ cargo loco task secret [length:20]
79e6760cc446d173a6eac736
```

## (2) As a integral `cargo loco secret` command

This has all of the advantages of "as a task", but also some *standardization* (the framework controls how the secret is generated and can also control the crypt hash and "push" best practices for security forward)

```
$ cargo loco secret [--length 20]
79e6760cc446d173a6eac736
```


## (3) Give the user some advice

What crypt to use, some unix command / etc. 

I feel option (2) allows us more control over the way secrets are generated so if we change things in the future it will be transparent for users

kaplanelad:
@jondot I think we can implement both aspects:
1. Establish a task for managing the JWT secret.
2. Develop a production YAML configuration in `cargo loco generate deployment` and give our best practices configuration for production service

jondot:
re (1) I feel that if we just "frame it" as a secret (not specific for JWT) it would be better, useful for different things (also for cookies in the future)

re (2) I think, we should provide a "best practice" yaml for production as part of the starters (for each starter), instead of empty that we have today. but we can have a lesson learned from Rails here as well -- for values that are arguable (people argue about / holy wars etc.) -- we leave it undefined, and make sure it fails appropriately with a series of checks, offering a way to resolve. For example:

```
$ LOCO_ENV=production cargo loco start

error: cannot start, because a JWT secret is not set in production.yaml. 

Did you forget to generate a secret?
run `cargo loco secret` to generate a new secret and set it in your `production.yaml`
```

What does this do? it makes all of these from argument where a user comes with "I don't agree" to "everything is OK" from our point of view:

1. People who don't trust our secret generator, will generate their own -- OK, because we gave them a clear error
2. People who do, will use our generator -- OK, because we gave them how to generate
3. People who don't want to set a secret in their yaml out of security, but want it in a vault and then pull it from ENV -- OK, because we didn't inject a secret automatically into their production yaml
4. In our best practice YAML, secret value should from from ENV by default -- OK, because we educate for "this should not be hardcoded", but they can disagree and edit it away

slyons:
I'm all for having a command that generates a `production.yaml` (or maybe one that prompts for the environment name), but what's the reasoning for having it as a task versus as a command on the CLI?

src/cli.rs
```diff
@@ -156,6 +156,8 @@ enum ComponentArg {
     },
     /// Generate a deployment infrastructure
     Deployment {},
+    /// Generate a new JWT secret
+    JwtSecret {}
```
kaplanelad, line 160: Let do it more generic and call it secret

```rust
  /// Generate a new  secret
  Secret {}
```

kaplanelad:
@slyons From what I see in Ruby, the secret command is not getting parameters in the CLI. 

I think we can keep the 20-char length static without the option to change it.

kaplanelad:
@slyons to pass the tests, you need to update the CLI snapshots
please read the guide [here](https://github.com/loco-rs/loco/blob/master/CONTRIBUTING.md#testing) how to do it

slyons:
@kaplanelad Even following the `TRYCMD` step it doesn't run the CLI tests for me:
```
test src/app.rs - app::Hooks::boot (line 79) ... ignored
test src/app.rs - app::Hooks::boot (line 86) ... ignored
test src/cli.rs - cli (line 7) ... ignored
test src/cli.rs - cli::main (line 254) ... ignored
```

kaplanelad:
@slyons you can update the `TRYCND` snapshot with the following command:
```
LOCO_CI_MODE=true TRYCMD=overwrite cargo test
```

slyons:
I'm getting the same result:
```shell
❯ LOCO_CI_MODE=true TRYCMD=overwrite cargo test
...
test src/app.rs - app::Hooks::boot (line 79) ... ignored
test src/app.rs - app::Hooks::boot (line 86) ... ignored
test src/cli.rs - cli (line 7) ... ignored
test src/cli.rs - cli::main (line 254) ... ignored
test src/app.rs - app::Hooks (line 52) ... ok
```

I've tried both in the `loco` root and in the `loco-cli` subcrate.

kaplanelad:
when you write `TRYCMD=overwrite cargo test --all-features --all` command, you don't see and changes in the test files?

Virviil:
@slyons 

Please, run the command
```
TRYCMD=overwrite cargo test --all-features --all
```
inside `examples/demo` folder in the project.

Also, ensure that you have valid connection to test database. By default it's configured like 

```uri: {{get_env(name="DATABASE_URL", default="postgres://loco:loco@localhost:5432/loco_app")}} ```

I'm using `DATABASE_URL="postgres://postgres:postgres@localhost:5432/postgres" cargo ...`

Also, no-one will create `DATABASE` for you, so ensure it exists.


It would be great if you'll add all these instructions to appropriate `*md* files inside this PR.

slyons:
@kaplanelad As I mentioned on the Discord, I've run the command in the example folder and still doesn't produce any changes. If someone else could just do it for me I'd appreciate it

PR #313, opened 2024-01-04
Added configuration and startup guide for mailer documentation

yinho999:
Added configuration and startup documentation with examples. 

Not sure do I need to add the commit in CHANGELOG.md since it is just documentation update.

### DIFF
--- a/docs-site/content/docs/the-app/mailers.md
+++ b/docs-site/content/docs/the-app/mailers.md
@@ -38,7 +38,7 @@ async fn register(
 This will enqueue a mail delivery job. The action is instant because the delivery will be performed later in the background.
 
 ## Mailer Configuration
-Configuration for mailers is done in the `config/[stage].toml` file. The default configuration is:
+Configuration for mailers is done in the `config/[stage].toml` file. Here is the default configuration: 
 
 ```toml
 # Mailer Configuration.
@@ -58,7 +58,7 @@ mailer:
     #   password:
 ```
 
-Mailer is done by sending emails to a SMTP server. An example configuration for using sendgrid (choosing the SMTP relay option) is:
+Mailer is done by sending emails to a SMTP server. An example configuration for using sendgrid (choosing the SMTP relay option):
 
 ```toml
 # Mailer Configuration.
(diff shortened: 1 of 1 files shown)


docs-site/content/docs/the-app/mailers.md
```diff
@@ -37,6 +37,63 @@ async fn register(
...
+    port: 465
+    # Use secure connection (SSL/TLS).
+    secure: true
+    auth:
+      user: "apikey"
+      password: "your-sendgrid-api-key"
+```
+
+### Default Email Address
+In addition to the SMTP configuration, you also need to configure the default email address that will be used as the sender for all emails. This email is most likely the email you registered with your email provider. You can set this by changing the `DEFAULT_FROM_SENDER` constant variable in the `src/mailer/mod.rs` file.
```
jondot, line 82: this is actually recommending to change the loco framework -- which the user cannot do.
the real fix for it is to override `opts` in your mailers:

```rust
   #[must_use]
    fn opts() -> MailerOpts {
        MailerOpts {
            from: DEFAULT_FROM_SENDER.to_string(),
            ..Default::default()
        }
    }
```

Here, you can supply your defaults as you wish instead of the stuff you get from the framework

yinho999: @jondot Thanks for your suggestions! However, isn't both code stored within the same file? 

``` rust
   // src/mail/mod.rs line 17
  pub const DEFAULT_FROM_SENDER: &str = "System <system@example.com>";
```
``` rust
   // src/mail/mod.rs line 59
    #[must_use]
    fn opts() -> MailerOpts {
        MailerOpts {
            from: DEFAULT_FROM_SENDER.to_string(),
            ..Default::default()
        }
    }
```

jondot: no i think one of those is being inside the framework
isn't it?

yinho999: @jondot Oh true, you are correct. both code you mentioned above are within the framework, I will update the documentation later. Just another question, MailerOpts is private by default. Should I make from and reply_to to be public for opts() function?

```rust
/// The options struct for configuring the email sender.
#[derive(Default)]
#[allow(clippy::module_name_repetitions)]
pub struct MailerOpts {
    from: String,
    reply_to: Option<String>,
}

#[async_trait]
pub trait Mailer {
   #[must_use]
    fn opts() -> MailerOpts {
        MailerOpts {
            from: DEFAULT_FROM_SENDER.to_string(),
            ..Default::default()
        }
    }
}
```

jondot:
did some editing to push forward, thanks for the docs! :heart:

yinho999:
> did some editing to push forward, thanks for the docs! :heart:

Sorry for the delay, was planning to finish it tonight 😁

Issue #307, opened 2024-01-03
Floats not supported in model generation


futuretim:
Unless I'm mistaken given the latest [schema](https://docs.rs/loco-rs/0.1.9/loco_rs/schema/index.html) f32 and f64 are not supported for mapping when generating a new model.

You can of course work around this by directly using SeaORM's ColumnDef::new directly instead and using the float function on it to set its type but this certainly breaks the usability for users because now it's a mixture of Loco and SeaORM.

I suspect this needs to be added? Please correct me if I'm wrong. Thanks!

jondot:
Yes, it can be added.
Would you like to work on it and submit a PR?

slyons:
I had to do this workaround myself. If I get time I might be able to add it

PR #317, opened 2024-01-05
Added float support in model generation

shuvroroy:
Issue: #307

### DIFF
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -2,7 +2,7 @@
 title = "Models"
 description = ""
 date = 2021-05-01T18:10:00+00:00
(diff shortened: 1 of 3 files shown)


futuretim:
Should we just go ahead and add double at the same time?

shuvroroy:
> Should we just go ahead and add double at the same time?

Waiting @kaplanelad review. If you approves then I will add

slyons:
I believe the original issue also mentioned `Date`

enieber:
nice job @shuvroroy...

enieber:
@slyons to add Date can be in another PR

kaplanelad (approved):
Thanks @shuvroroy

Issue #325, opened 2024-01-07
Document how to work with `ApiToken`


kaplanelad:
Loco offers two middleware options for authentication:

1. Authentication based on a username and password, with detailed documentation available [here](https://loco.rs/docs/features/authentication/#get-current-user).

2. Authentication based on a user API token, which is currently not documented on loco.rs.

Needs to document the authentication method based on API tokens. The relevant code reference is in the [middleware implementation](https://github.com/loco-rs/loco/blob/master/src/controller/middleware/auth.rs#L140).

For a practical example of implementing this middleware, refer to the following code snippet in the [user controller](https://github.com/loco-rs/loco/blob/master/examples/demo/src/controllers/user.rs#L13).

yuki0418:
@kaplanelad 
I'm a new so can try this.

kaplanelad:
Hey @yuki0418, no problem at all! I'd be happy to assist you, and we can tackle this together.

Let's follow these steps:

1. Begin by creating a new project using `loco new`. Choose either saas or rest when prompted.
2. Register a new user by referring to the documentation [here](https://loco.rs/docs/features/authentication/#registering-a-new-user).
3. Once you've made the request, check the database to ensure that the user you created has an api_token stored.
4. Next, create a new endpoint similar to this [example](https://github.com/loco-rs/loco/blob/master/examples/demo/src/controllers/user.rs#L13), configuring the auth middleware with the API token.

Once you've successfully navigated through these steps, you're ready to start creating the documentation :) Feel free to reach out if you encounter any issues.

yuki0418:
@kaplanelad 
Thank you so much for leading me!!

I check current user controller API code in demo and it uses `auth::JWTWithUser<users::Model>` for `current fn` [here,](https://github.com/loco-rs/loco/blob/006f532b0cef2ff0838ca7c0f9e40a66fe7f1a3a/examples/demo/src/controllers/user.rs#L6C1-L6C1) however, the current doc([here](https://github.com/loco-rs/loco/blob/006f532b0cef2ff0838ca7c0f9e40a66fe7f1a3a/docs-site/content/docs/features/authentication.md?plain=1#L168)) shows `auth: middleware::auth::Auth,`.
I guess the doc is old, so I will fix it too. Tell me if it not necessary. 

(user controllers in staters use still `auth::JWT`. I assume it is just before fix.)

yuki0418:
Now I successfully checked new `ApiToken` middleware, so move on fix doc.

[...]

PR #328, opened 2024-01-07
Update authentication.md for ApiToken middleware

yuki0418:
## Changes
- Add ApiToken middleware example into authentication.md.
- Change old api `middleware::auth::Auth` to `middleware::auth::JWT` in doc
- Update docs-site/content/blog/hello-world.md

### DIFF
--- a/docs-site/content/docs/features/authentication.md
+++ b/docs-site/content/docs/features/authentication.md
@@ -175,3 +175,89 @@ async fn current(
 }
 
 ```
+
+## API Authentication
+
+### Creating new app
+
+For this time, let create your rest app using the [loco-cli](/docs/getting-started/tour) and select the `Rest app` option.
+To create new app, run the following command and follow the instructions:
+
+```sh
+$ loco new
+```
+
+To explore the out-of-the-box auth controllers, run the following command:
+
+```sh
+$ cargo loco routes
+ .
+ .
+ .
+[POST] /api/auth/forgot
+[POST] /api/auth/login
+[POST] /api/auth/register
+[POST] /api/auth/reset
+[POST] /api/auth/verify
+[GET] /api/user/current
+ .
+ .
+ .
+```
+
+### Registering new user
+
+The `/api/auth/register` endpoint creates a new user in the database with an `api_key` for request authentication. `api_key` will be used for authentication in the future requests.
+
+#### Example Curl Request:
+
+```sh
+curl --location '127.0.0.1:3000/api/auth/register' \
+     --header 'Content-Type: application/json' \
+     --data-raw '{
+         "name": "Loco user",
+         "email": "user@loco.rs",
+         "password": "12341234"
+     }'
+```
+
+After registering a new user, make sure you see the `api_key` in the database for the new user.
+
+### Creating an Authenticated Endpoint with API Authentication
+
+To set up an API-authenticated endpoint, import `controller::middleware` from the loco_rs library and include the auth middleware in the function endpoint parameters using `middleware::auth::ApiToken`.
+
+Consider the following example in Rust:
+
+```rust
+use loco_rs::prelude::*;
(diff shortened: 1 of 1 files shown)


docs-site/content/docs/features/authentication.md
```diff
@@ -163,13 +164,32 @@ use loco_rs::{
...
+
+/**
+ * Authorization header should have token which is generated email and password login function.
+ * 
+ * HTTP Headers:
+ * Authorization: Bearer TOKEN
+ */
 async fn current(
-    auth: middleware::auth::Auth,
+    auth: middleware::auth::JWT,
```
yuki0418, line 176: I think we don't use Auth anymore. So fix here as well.
Tell me if I'm wrong.

kaplanelad:
@yuki0418 
The idea is to add another doc section that describes how to use the API middleware and not to change the current docs. 

As I wrote you in the issue, we have two middleware. The first one is an auth-based user and password with works and documented. 
The second is the auth based on API that needs to be documented. i think you can add another  section in the auth docs that describes how to use the API auth middleware

kaplanelad:
@yuki0418 yes, adding a new section is great! 

I think you can elaborate more on the flow to the readers as I wrote you in the [issue ](https://github.com/loco-rs/loco/issues/325#issuecomment-1880028885)

I see the changes like that.
1. On top of the file example, we provide middleware for authentication in `Loco simplifies the user...`
2. Move all the current doc to a new section called user password authentication
3. Then create a new section called API authentication that explain the flow like I wrote you in the issue + curl command example 

WDYT?

yuki0418:
@kaplanelad 
Update doc, please check it.

image

docs-site/content/docs/features/authentication.md
```diff
@@ -14,15 +14,17 @@ toc = true
...
-## Authentication Configuration
+### Authentication Configuration
 
 The `auth` feature comes as a default with the library. If desired, you can turn it off and handle authentication manually.
 
-## Getting Started with a Saas App
+### Getting Started with a Saas App
 
-Create your app using the [loco-cli](/docs/getting-started/tour.md) and select the `Saas app` option.
+Create your app using the [loco-cli](/docs/getting-started/tour) and select the `Saas app` option.
```
yuki0418, line 27: URL is shows 404. Fix it.

docs-site/content/docs/features/authentication.md
```diff
@@ -173,3 +175,89 @@ async fn current(
 }
 
 ```
+
+## API Authentication
```
yuki0418, line 179: The title `API Based Authentication` would be better?

docs-site/content/docs/features/authentication.md
kaplanelad: Does this link work for you when you run the app?
i think you should replace it to 
[loco-cli](@/docs/getting-started/tour.md)

kaplanelad: Does this link work for you when you run the app?
i think you should replace it to 
[loco-cli](@/docs/getting-started/tour.md)

kaplanelad (changes requested):

docs-site/content/docs/features/authentication.md
yuki0418: @kaplanelad 
Yes works when run on dev with `zola serve`
I attach SS of dev and prod below.

The reason of it because now `docs-site/content/docs/getting-started/tour/index.md` so there is no `tour.md` anymore.

## Fixed
https://github.com/loco-rs/loco/assets/45615348/7c439d24-112d-4b73-b49d-ce3add91d01f

And I check prod, it NOT works.

https://github.com/loco-rs/loco/assets/45615348/893e7651-3c57-4313-adc5-d00c9179efdc

kaplanelad (approved):

Issue #221, opened 2023-12-22
Document: default fields in generated model


slyons:
**Description**

I have a bug in one of my migrations (that's another issue entirely), but it's causing `cargo loco db migrate` to fail silently.

**To Reproduce**

Starting from the REST API template:
`cargo loco generate model items name:string! description:string quantity:int! size:int! infinite:bool! created_at:ts! user:references`

```shell
❯ cargo loco db migrate
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/roadiebag3-cli db migrate`
2023-12-22T23:09:37.786766Z  WARN loco_rs::boot: pretty backtraces are enabled (this is great for development but has a runtime cost for production. disable with `logger.pretty_backtrace` in your config yaml)
2023-12-22T23:09:37.910313Z  WARN loco_rs::boot: migrate:
2023-12-22T23:09:37.914611Z  INFO sea_orm_migration::migrator: Applying all pending migrations
2023-12-22T23:09:37.919476Z  INFO sea_orm_migration::migrator: Applying migration 'm20220101_000001_users'
2023-12-22T23:09:37.930024Z  INFO sea_orm_migration::migrator: Migration 'm20220101_000001_users' has been applied
2023-12-22T23:09:37.931591Z  INFO sea_orm_migration::migrator: Applying migration 'm20231103_114510_notes'
2023-12-22T23:09:37.935946Z  INFO sea_orm_migration::migrator: Migration 'm20231103_114510_notes' has been applied
2023-12-22T23:09:37.936514Z  INFO sea_orm_migration::migrator: Applying migration 'm20231222_225942_items'
```
The problem is I have one more model after `items`, and so the migration never runs.

But if I run `cd migration && cargo run` I get:

```
Applying all pending migrations
Applying migration 'm20220101_000001_users'
Migration 'm20220101_000001_users' has been applied
Applying migration 'm20231103_114510_notes'
Migration 'm20231103_114510_notes' has been applied
Applying migration 'm20231222_225942_items'
Execution Error: error returned from database: column "created_at" specified more than once
```

**Expected Behavior**

The migrations are to run so I can generate my entities.

**Environment:**

Local development, trying to follow the guide.

**Additional Context**

jondot:
Thanks!
FYI - When generating a model you don't need any timestamps (we give you created_at and updated_at for free, similar to what Rails does)

[...]

PR #335, opened 2024-01-08
Document migration default values

OdujokoD:
This PR addresses #221

### DIFF
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -72,10 +72,12 @@ To add a new model _you have to use a migration_.
 $ cargo loco generate model posts title:string! content:text user:references
 ```
 
-When a model is added via migration, the following default values are provided:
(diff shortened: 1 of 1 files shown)


docs-site/content/docs/the-app/models.md
```diff
@@ -72,6 +72,11 @@ To add a new model _you have to use a migration_.
 $ cargo loco generate model posts title:string! content:text user:references
 ```
 
+When a model is added via migration, the following default values are provided:
```
kaplanelad, line 75: Thank you, @OdujokoD.

If the user includes fields such as `created_at`,`updated_at`,`create_at, or `update_at`, we are currently disregarding those specific fields. You can see the code [here](https://github.com/loco-rs/loco/blob/f0b4166f2e7678d50e4044e5b5ba3988987c0a23/src/gen/model.rs#L20).

Could you also document this behavior, please?

kaplanelad (approved):

PR #337, opened 2024-01-09
Add an `Enabled<T>` type that allows turning configuration off without having to specify all fields.

felipesere:
This PR shows a technique to allow disabled sections of the configuration to not have to list all fields, potentially with bogus values.

For example, given:
```rust
struct WithStaticAssets {
  enable: bool,
  path: String,
  size: u32,
}
```
would deserialize from, assuming some other struct `X` to make the indentation nice:
```yaml
x:
  enable: true
  path: /path/to/somewhere
  size: 128
```
That is fine, but what about when we want to turn it off? We still need all required fields:
```yaml
x:
  enable: false
  path: /path/to/somewhere/that/is/ignored
  size: 0
```

With `Enabled<T>` the struct would become
```rust
struct WithStaticAssets {
  path: String,
  size: u32,
}

struct X {
  x: Enabled<WithStaticAssets>,
}
```

which then supports all of the following cases.

* enabled
```yaml
x:
  enabled: true
  path: /path/to/somewhere
  size: 128
```
* disabled with extra fields:
```yaml
x:
  enabled: false
  path: /path/to/somewhere
  size: 128
```

* disabled without any extras
```yaml
x:
  enabled: false
```

I've added some extra methods like `is_enabled()` just to show how to possible work with it.


felipesere:
@kaplanelad 👋 any opinions on this suggestion? 😄

jondot:
@felipesere would this break existing apps with configurations? (assuming they always have all fields specified either way)

felipesere:
Not unless I've fumbled `enable` vs `enabled`? As the extra fields are ignored when the state is `enabled: false`.

jondot:
Ive activated CI
I think if this is non-breaking it can be an interesting upgrade to how the "enabled/disabled" mechanism works

jondot:
Im also wondering -- can this mechanism be extracted to a separate library? it looks like a nicer helper library for Serde, and the code itself seems non-related to our specific Config logic, so might be a win-win here.

felipesere:
Yeah I was thinking it might be a nice little library!

jondot:
Cool, is there a chance you extract it and then pull it back to this PR? we can wait, theres no rush

felipesere:
I have extracted it into a library, but haven't published it yet. 
You can see the source here: https://github.com/felipesere/serde_enabled

I think the types are a bit nicer, since it comes down to
```rust
pub enum Enabled<T> {
  On(T),
  Off
}
```
without exposing the extra inner ones 😅

felipesere:
Added a the new library. Give it a whirl!

src/cli.rs
```diff
@@ -264,9 +264,8 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
     let environment: Environment = cli.environment.unwrap_or_else(resolve_from_env).into();
 
     let config = environment.load()?;
-
-    if !H::init_logger(&config, &environment)? {
-        logger::init::<H>(&config.logger);
+    if let Some(l) = config.logger.as_ref() {
```
jondot, line 267: Trying to discuss this for a minute, so we're saying - if there's no logger section, logging is never initialized?

I'm torn about that. Somehow I feel there should always be a logger section and then control it with enable/disable. I don't know why but logging feels like a special case where you always want to see that config visually and then tweak it but never completely remove it from a file.

WDYT?

src/controller/app_routes.rs
```diff
@@ -177,49 +177,69 @@ impl AppRoutes {
...
-        if let Some(limit) = &ctx.config.server.middlewares.limit_payload {
-            if limit.enable {
-                app = Self::add_limit_payload_middleware(app, limit)?;
-            }
+        if let Some(limit) = ctx
+            .config
+            .server
+            .middlewares
+            .limit_payload
+            .clone()
```
jondot, line 196: any way to shorten this chain? I'm not sure myself just wondering, because if it ends up in a ref eventually, maybe we can supply a ref directly

src/controller/app_routes.rs
```diff
@@ -177,49 +177,69 @@ impl AppRoutes {
...
+            .and_then(|l| l.into_inner())
+        {
+            app = Self::add_limit_payload_middleware(app, &limit)?;
         }
 
         if let Some(logger) = &ctx.config.server.middlewares.logger {
-            if logger.enable {
+            if logger.is_enabled() {
```
jondot, line 203: i'm wondering why the structure here is different from the other middleware powered by the new Enable library

src/cli.rs
felipesere: Yeah this is a fair point.

I think this warrants adding some tests as exemplars 😅

src/controller/app_routes.rs
felipesere: Excellent catch. I've tidied my approach a buch here in a recent commit.

src/controller/app_routes.rs
felipesere: Jup. A little at least. Its a bit annoying that it is `Option<Enable<...>>` which means I have to poke through those containers a little.

src/cli.rs
felipesere: I've changed this now in c85dde732ab3c7ee0d16498d7b818ad142ffb5f5 to use a default logger config when its either not configured `None` or disabled `Some(Enabled::Off)`.

The fields can still be present in the config file, even when the feature is `Enabled::Off`. 
They are just not required 😅

felipesere:
@jondot I've rebased this with `main` and noticed a breaking change in the config. The addition of `precompressed` without a default fails serialisation.

I wonder if as an initial PR I should just introduce a bunch of tests for config setups?
And then work `Enabled<T>` back in? It was pretty annoying that the Error was not obvious, so I'm thinking of incorporating https://github.com/dtolnay/path-to-error into my side of the library. What do you think?

The remaining issues _should_ be resolved 😬

jondot:
Appreciate the rebase effort!
path to error looks interesting, I can't help but think if we can incorporate it into all serialization errors in Loco.
what do you think? is there a downside to it?

felipesere:
I don't fully grok the 2 failing tests though. Could you maybe take a look?

jondot:
The failure says something like: "somewhere in your code, you have initialized tracing twice". it's probably the logger logic, it is being initialized once and then second init creates a panic.
This is a known limitation of the tracing crate btw, we have no chance but to accept this limitation and always hunt for that double-initialization

felipesere:
This has totally fallen by the wayside for me. I'll see if I can get the branch updated and then those errors fixed by Sunday 🤞

felipesere:
@jondot is there a way I can trigger these builds myself? on my my own fork/branch? Should I be copying the workflows files?

kepae:
(from left field!) @felipesere you can indeed run the GitHub workflows configured in this repository in your own fork if you open a PR inside there. A little redundant but since it's trigger on branch commit, both your fork and the upstream PRs (if open) will trigger the actions.
I just did the same in mine for a simple fix: https://github.com/kepae/loco/pull/1
The files configuring the actions are included in your fork: https://github.com/felipesere/loco/tree/master/.github/workflows

felipesere:
Let me know if you are keen to revisit this 😅

jondot:
Hi, since we've last discussed, we've finalized a reimplementation of the middleware configuration settings (in v0.10.0 that was released today), it should make things easier so closing this for now.

Issue #342, opened 2024-01-10
Add HTMX Support


ryankopf:
## Feature Request

HTMX is great and hugely up-and-coming in the frontend world. HTMX is a powerful tool that allows us to enrich our web applications with dynamic content, without the complexity of a full-blown JavaScript framework. This addition would streamline the development process, significantly reduce the learning curve for new developers, and enhance the overall user experience.

Key Benefits:
- Very "Railsy" - HTMX is a lot like turbo and Hotwire, except its not currently just limited to Rails. Integration will be relatively simple.
- Ease of Use: HTMX's syntax is intuitive and aligns well with our existing HTML-centric development approach. It simplifies the process of adding interactive features to web pages.
- Improved Performance: By using HTMX, we can minimize the amount of JavaScript needed, leading to faster page loads and a smoother user experience, especially on lower-powered devices.
- Enhanced Scalability: HTMX's lightweight nature and compatibility with standard HTTP methods make it a scalable solution for both small and large-scale applications.
- Seamless Integration: Given its design, HTMX can be seamlessly integrated with the existing Rust architecture, maintaining the robustness and reliability we've established.
- Reduced Complexity: It reduces the complexity of our codebase by replacing intricate JavaScript functionalities with simple HTML extensions, making it easier to maintain and debug.

jondot:
Thanks @ryankopf 
As much as I was able to gather, htmx is indeed closest to what Rails has today with Turbo.
Would you like to submit some draft or PR that showcases how it all should work?

ryankopf:
I've created a PR that is necessary to start on this. https://github.com/loco-rs/loco/pull/344

My PR renames "SaaS App" to "React Frontend". This is necessary so that when I go to create the HTMX frontend, it will make idiomatic sense what all the options are.

In the future, imagine generating a project will look like this:

```
What would you like to build?
> Stateless App
> API-Only App
> React Frontend
> HTMX Frontend
```

jondot:
Thanks Ryan ❤️ , moving this to discussion (after I reviewed the PR)

PR #344, opened 2024-01-11
Rename SaaS frontend to React frontend

ryankopf:
As we should be supporting any number of frontends based on what developers want, each frontend should have a name reflecting the technology it uses.

This PR renames "SaaS App" to "React App" in all places it is used. Because in the future, the options will probably look like:
```
> Stateless App
> API-Only App
> React Frontend
> HTMX Frontend
> Other frontends as desired
```

Since I hope HTMX is worked on from [this issue](https://github.com/loco-rs/loco/issues/342), this is a necessary first step before we can approach HTMX.

I'm not quite sure how to test everything properly, but I did the following test and debugging steps to ensure this works. If I am missing any steps, let me know. Please also confirm that everything works as expected from your experience as well.

In the root project folder:
```
cargo test
```
Everything passes except 1 failure, but the failure is due to windows line endings in the mailer template "/r/n" so the test just needs to be updated later to accommodate either.

Then in the /loco-cli/ folder

```
cargo test
```

and all tests pass!

Here's the tricky part - the starters are generated from the BASE_REPO_URL, so I had to change that in my code while testing this to actually get the updated starters.

I had to modify "BASE_REPO_URL" to target my updated branch, and so I also had to specify a BRANCH_NAME. I figure this feature is useful to retain for future developers as they may add more frontends, making this easier to debug.

So in my case, to run tests, I had to set:
```
const BASE_REPO_URL: &str = "https://github.com/ryankopf/loco.git";
const BRANCH_NAME: &str = "react-frontend";
```

cargo build
.\target\debug\loco new
Presents the new options!


kaplanelad:
Your efforts are genuinely commendable; thank you for the extensive work you've put in.

We should not support all kinds of started (like you write in the description).
Many starters mean a lot of maintenance and work and depend on a lot of changes in many places. the start should be minimalist, small and contain a context,

As I see it adding all these kinds of templates should be like [this](https://github.com/loco-rs/todo-list) repo that people can clone and start to work on.

ryankopf:
(Closed PR on accident, sorry.)

ryankopf:
Further note, I separated the PRs renaming "SaaS App" and creating the "HTMX Frontend" into two different PRs for exactly this reason.

Even if we don't support additional frontends, "SaaS App" is misleading when it is really "API + React Frontend".

kaplanelad:
@ryankopf, as mentioned in my previous comment, For now, I prefer not to open the starters selection per use case (e.g., SaaS app, SaaS with React, SaaS with HTMX, etc.).

I suggest creating a project similar to [this one](https://github.com/loco-rs/todo-list). Users can clone it and start their work from there.

ryankopf:
@kaplanelad I think the community should have a say regarding which starters are included. All the guides say to use "cargo loco new" to create a project, and having a "SaaS App" that starts with React is an unexpected experience for many devs. Further, a framework SHOULD be as JavaScript-agnostic as possible, which is why the recommendation for HTMX; it would be much easier to replace than replacing React if the default frontend is React. As I've said, I'm willing to volunteer to maintain an official HTMX starter.

Alternatively, all starters should be removed from the project and we should change the docs to suggest cloning one of the examples as you proposed.

kaplanelad:
Reconsidering the way we serve starters is a valid idea. However, I suggest implementing such changes after we establish a clear vision and engage in thorough discussions about these modifications because this will affect on how we going to maintain and manage those apps.

So, at this point, I recommend opening a discussion, making a decision there, and how it fits into the roadmap of @jondot and I

kaplanelad:
@ryankopf, I am closing this pull request for now.

Issue #359, opened 2024-01-16
Documentation issues with Tour and Loco Guide


KnowZero:
https://loco.rs/docs/getting-started/tour/

When you get to the "Checking Out SaaS Authentication" section, it says nothing about starting up Redis docker so you get a bb8 timeout error

Also, the example mostly sets up a backend for a blog, with no front end. It probably should be clear that it is a backend of a blog so someone doesn't think something isn't working when they don't see a full blog

https://loco.rs/docs/getting-started/guide/

To test the examples, I created myapp_guide in the same folder so it doesn't conflict with the above folder

When I did `cargo loco doctor`, it gives this error:
```
error: unrecognized subcommand 'doctor'

Usage: myapp_guide-cli [OPTIONS] <COMMAND>

For more information, try '--help'.
```
Note: In the example of the blog backend, it works fine

curl localhost:3000/api/_ping
and
curl localhost:3000/api/_health

also do not work, and like above they work for the blog backend example

If I do `cargo loco routes`, it is missing from the routes:
[GET] /api

PS It is not clear if documentation issues should be posted here, discussions, or some other place, sorry if this is the wrong place

kaplanelad:
Hey @KnowZero
about the Redis documentation, do you want to open a pull request with this addition?

for the `cargo loco doctor` which starter are you selected? can you share with me the `--help` output?

KnowZero:
Sure, I can do that for the Redis. What about changing blog to "blog backend" part? Like so:

Let's create a blog on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:

to

Let's create a blog backend on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:

Should I do that too?

For the starter, I picked the one mentioned in the guide "lightweight-service (minimal, only controllers and views)"

Here is Help:
```
myapp_guide → cargo loco help
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/myapp_guide-cli help`
The one-person framework for Rust

Usage: myapp_guide-cli [OPTIONS] <COMMAND>

Commands:
  start     Start an app
  routes    Describe all application endpoints
  task      Run a custom task
  generate  code generation creates a set of files and code templates based on a predefined set of rules
  version   Display the app version
  help      Print this message or the help of the given subcommand(s)

Options:
  -e, --environment <ENVIRONMENT>  Specify the environment [default: development]
  -h, --help                       Print help
[...]

PR #360, opened 2024-01-16
Site doc fixes for #359

KnowZero:
This commit includes the following fixes to the site documentation:

1. Mentions you need to run redis in the Quick Tour
2. Clarifies that the Quick Tour creates a site backend
3. Updates the Quick Tour `loco new` text to reflect current text
4. Updates the guide `loco new` text to clearer show the starter choice with an arrow

### DIFF
--- a/docs-site/content/docs/getting-started/tour/index.md
+++ b/docs-site/content/docs/getting-started/tour/index.md
@@ -16,7 +16,7 @@ top = false
 <br/>
 <br/>
 <br/>
-Let's create a backend on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:
+Let's create a blog backend on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:
 
 ```sh
 $ cargo install loco-cli
--- a/docs-site/content/docs/getting-started/tour/index.md
+++ b/docs-site/content/docs/getting-started/tour/index.md
@@ -16,7 +16,7 @@ top = false
 <br/>
 <br/>
 <br/>
-Let's create a blog on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:
+Let's create a backend on `loco` in 4 commands. First install `loco-cli` and `sea-orm-cli`:
 
 ```sh
 $ cargo install loco-cli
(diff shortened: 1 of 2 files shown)


docs-site/content/docs/getting-started/tour/index.md
```diff
@@ -140,6 +142,22 @@ Done! enjoy your ride with `loco` 🚂
 
 Your generated app contains a fully working authentication suite, based on JWTs.
 
+To authenticate, you will need a running redis server.
```
jondot, line 145: To make the tutorial with less friction, we can switch the worker mode to inprocess background: `BackgroundAsync`.
This removes the need for Redis.
The need for workers is for background sending of registration emails etc.


Reminder, the worker config:

```
 # Worker Configuration
workers:
  # specifies the worker mode. Options:
  #   - BackgroundQueue - Workers operate asynchronously in the background, processing queued.
  #   - ForegroundBlocking - Workers operate in the foreground and block until tasks are completed.
  #   - BackgroundAsync - Workers operate asynchronously in the background, processing tasks with async capabilities.
  mode: BackgroundQueue

```

KnowZero: So how to go about this? Change the Saas default to BackgroundAsync or have them manually change it in the Tour?

Also, if BackgroundAsync is inprocess, then the config documentation should be clear about which ones require external database and which ones are done inprocess so I'll update that as well I guess

jondot: Good question. 
We're asking the user to do either of the two:

1. Configure and set up Redis
2. Change a setting in their config.yaml

What do you think makes the least friction and mental effort? (we want to remember that this is kind of a "quick tour")

KnowZero: If we are talking about most easiest, since you are already setting up docker for Postgres and are already on the shell, copy and pasting running Redis is probably easier than going through a config file. Unless you are setting that config file via shell as well. I mean there is nothing less mental effort than copy and paste

On the up side, changing the config file gets the user more aware of the config files existance

jondot: yes, that sounds convincing. so let's go with bringing up another Docker.

PR #379, opened 2024-01-20
issue-361 added a binding flag for specifying server listen address

mikosco4real:
- Added a cli optional argument for binding address when starting loco server
- Updated the documentation to reflect this change

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -32,6 +32,7 @@ use clap::{Parser, Subcommand};
 use crate::{
     app::{AppContext, Hooks},
     boot::{create_app, create_context, list_endpoints, run_task, start, RunDbCommand, StartMode},
+    config::ConfigOverrides,
     environment::{resolve_from_env, Environment, DEFAULT_ENVIRONMENT},
     gen::{self, Component},
     logger, Result,
@@ -232,8 +233,9 @@ where
 pub async fn playground<H: Hooks>() -> Result<AppContext> {
     let cli = Playground::parse();
     let environment: Environment = cli.environment.unwrap_or_else(resolve_from_env).into();
+    let config_overrides = ConfigOverrides::default();
 
-    let app_context = create_context::<H>(&environment).await?;
+    let app_context = create_context::<H>(&environment, &config_overrides).await?;
     Ok(app_context)
 }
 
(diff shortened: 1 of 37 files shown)


src/cli.rs
```diff
@@ -282,7 +286,11 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
                 StartMode::ServerOnly
             };
 
-            let boot_result = create_app::<H, M>(start_mode, &environment).await?;
+            let b = binding.unwrap_or_else(|| "[::]".to_string());
```
kaplanelad, line 289: Can you change variable `b` to `server_binding`?

src/cli.rs
```diff
@@ -282,7 +286,11 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
...
+
+            let mut boot_result = create_app::<H, M>(start_mode, &environment).await?;
```
kaplanelad, line 291: I would suggest that you refrain from mutating`boot_result` directly.  I think we can improve the `environment. Load ()?` function (see at the top of the function) by add a struct called `config_overrides` for examples. when loading the configuration, we can override the YAML configuration with the struct values.

kaplanelad: One more note in this file, 
We have 2 `main` functions. One is when the feature `with-db` is on, and one is off. Please make sure that we are supporting the `binding` flag in both of the methods

src/config.rs
```diff
@@ -256,6 +259,15 @@ impl Server {
...
         format!("{}:{}", self.host, self.port)
     }
+
+    pub fn add_binding_addr(&mut self, addr: &str) {
+        self.binding = Some(addr.to_string());
+    }
+
+    #[must_use]
+    pub fn get_binding(&self) -> String {
+        self.binding.clone().unwrap_or_else(|| "[::]".to_string())
```
kaplanelad, line 269: `[::]` is configured here and also in the `src/cli.ts.` take the `[::]` value from the cost (see previous comment)

kaplanelad: We can use  #[serde(default = "default_binding")].
 In the `Server` struct, you can replace the `pub binding` with this:
```rust
    #[serde(default = "default_binding")]
    resource: String,
```

And replace this function with this:
```rust
fn default_resource() -> String {
    "[::]".to_string()
}
```

This is allows us to avoid the `mut` in the `src/cli.rs` and also avoid `[::]` duplication default declaration

kaplanelad (changes requested):

src/config.rs
jondot: @kaplanelad  `resource` you mean `binding` right?

kaplanelad: yea sure, like i wrote on the field struct default_binding.

thanks for correct in the example

src/cli.rs
mikosco4real: @kaplanelad, I have made all the requested changes save for one. That is a way to avoid mutating `boot_result` directly. Can you explain in more details how you think I should approach this. Are you suggesting that I should mutate the environment instead. Also, in `cli.rs` the `environment.load()` is called before matching the CLI command which means that I will need to override the environment before then should I want to avoid mutating the environment.

kaplanelad: @mikosco4real, the primary objective is to avoid using mutable references on the `boot_result`. To achieve this, we propose shifting the decision-making process for overrides to the `configuration` and refrain from modifying the configuration already loaded within the `boot_result` struct.

To implement this, I suggest introducing a new struct named `ConfigOverrides`. This struct will be initialized with variables from the CLI, specifically, in this case, with a `server_binding` option:
```rust
pub struct ConfigOverrides {
   server_binding: Option<String>,
}
```

During the parsing of CLI arguments, set the binding flag into an instance of this struct. Subsequently, when loading the environment, pass this struct as an argument to the configuration loading process to effect the necessary overrides:
```rust
let config_overrides = ConfigOverrides {
    server_binding: /* CLI binding value */,
};

let config = environment.with_overrides(&config_overrides)?;
```

This approach allows for a more structured and modular handling of configuration overrides without the need for mutability on the boot_result.

kaplanelad:
Hello @mikosco4real,

I appreciate the adjustments you made. However, upon passing the config overrides to create_context, it appears to be something we'd rather steer clear of as well.

I took tp address this myself, and you can review the outcomes here: https://github.com/loco-rs/loco/pull/402

Thanks again for your collaboration.

Issue #329, opened 2024-01-07
Add an integrated template rendering stack + i18n


alexhumphreys:
## Feature Request

I couldn't find anything in the repo/roadmap in terms of internationalization (i18n). As i18n is usually easier to have from the start rather then to add later, it would be cool if there was some form of loco support for i18n out of the box. 

**Describe the solution you'd like**

No strong preference on "the solution I'd like". The usual basic solution for this is abstracting out the strings the application produces, and looking up the values in different languages. Same for things like date/currency formats that change between regions. Can check out the [Rails Internationalization (I18n) API](https://guides.rubyonrails.org/i18n.html) to see how they solve it.

**Describe alternatives you've considered**

Looks like loco is using [`tera`](https://docs.rs/tera/latest/tera/) for templating. There's this [`fluent-templates`](https://github.com/XAMPPRocky/fluent-templates#tera) crate which provides integration between the [fluent localization framework](https://projectfluent.org/) and tera. Maybe that or something similar could work?

PS. thanks for all the work on Loco, it's pretty sweet and just the kind of thing I've been looking for as I get into Rust ❤️

jondot:
Thanks Alex!

Do you see the i18n infra available through Tera? meaning this is the place where you need i18n support?

alexhumphreys:
Suppose it could be needed anywhere there's a string that can be shown to an end user in a different region. The two places I can think of like that are in templates, and error message strings, but I'm maybe missing some other places.

Guess it comes up a bit more in server rendered HTML apps, rather than apps just exposing pure json APIs, so guess it depends on exactly what usecases you have in mind for loco

jondot:
To make it a bit more actionable I guess we can start with integrating fluent-templates into our Tera support. Would you want to create a PR for it?

alexhumphreys:
Hmm before I open try get something in a PR, any idea what you'd like the interface to this to be? 

It looks like the three parts involved in using fluent-templates are:
- [import the locale directory](https://github.com/XAMPPRocky/fluent-templates/blob/master/README.md?plain=1#L184)
- [register a `fluent` function with tera, and pass the locale info as an argument](https://github.com/XAMPPRocky/fluent-templates/blob/master/README.md?plain=1#L197)
- [call the render function, while using the `fluent` function in your template](https://github.com/XAMPPRocky/fluent-templates/blob/master/README.md?plain=1#L200)

[...]

PR #385, opened 2024-01-22
Add fluent-template for template rendering stack + i18n 

alexhumphreys:
Opening this as a rough draft as part of #329 just to get some feedback. Not sure:
- how any of this should be named
- how tests should be done
- how config should be documented
- how errors should be handled
- what else I've missed

So there's more work to be done here 😄 But figured I'd open up a draft PR and start getting input before I go way off in the wrong direction.


jondot:
This looks good for assessing the effort required, looks achievable!

I can give you a few pointers:

* Start working from `examples/demo`, this will help you place assets (locales, templates) in a non-framework source tree, and then, once you have something working, think about "how do i extract functionality into the framework"
* Extracting into the framework means pieces of code that would be generic in their behavior (nothing specific to a specific app), and also -- picking the right name, module name, source file names, is important (we can help with that)

jondot:
I figured creating a PR of what I mean would be easier (managed to remove a few panics/edge cases, etc. in the original code too, so review the code itself too):

https://github.com/loco-rs/loco/pull/389

let's move to work there, I also made sure to document all the design questions, etc. so we can get your thoughts there as well

alexhumphreys:
@jondot ah that's awesome! Will close this so and move over there. Nice one!

PR #402, opened 2024-01-28
support binding and port in cli command

kaplanelad:


src/cli.rs
```diff
@@ -283,7 +294,14 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
             };
 
             let boot_result = create_app::<H, M>(start_mode, &environment).await?;
-            start(boot_result).await?;
+            let serve_config = ServeConfig {
+                port: port.map_or(boot_result.app_context.config.server.port, |p| p),
```
jondot, line 298: is it just `.or(...)`? (since the map is identity)

src/cli.rs
```diff
@@ -283,7 +294,14 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> eyre::Result<()> {
             };
 
             let boot_result = create_app::<H, M>(start_mode, &environment).await?;
-            start(boot_result).await?;
+            let serve_config = ServeConfig {
```
jondot, line 297: maybe rename to `ServeParams` so we know it didn't come directly from config?

mikosco4real:
Nicely done @kaplanelad, I did not think about approaching it this way. Always something new to learn. I think what's left is to update loco docs so users can know how to use the newly added flags to manage their server address and port.

PR #406, opened 2024-01-31
[WIP] Add paperclip like libarary

yinho999:
Working on file upload functionality built on [object store](https://crates.io/crates/object_store) which supports AWS S3, Azure, Google Cloud storage, in memory, Local storage and Http.

Current Implementation
1. Storage into AppContext.
2. Configurations for different variation of storage.

Todos
1. Testing for each storage,
2. Endpoints.
3. Documentation
4. Features Flags to enable each storage
5. Take in env configurations files

@kaplanelad would you mind having a look at this, please?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ testing = ["dep:axum-test"]
 with-db = ["dep:sea-orm", "dep:sea-orm-migration"]
 channels = ["dep:socketioxide"]
 
+store-aws = ["object_store/aws"]
+store-gcp = ["object_store/gcp"]
+store-azure = ["object_store/azure"]
 
 [dependencies]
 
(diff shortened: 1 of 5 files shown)


yinho999:
FYI the code in this branch is not fully working, the example code will be as the following for local storage and S3 bucket
https://github.com/yinho999/loco-playground/blob/master/src/bin/multipart.rs

kaplanelad:
Hey @yinho999 
Upon further exploration of this implementation, I believe adopting your suggestion and utilizing the object_store library is the way to go. This library offers enhanced integration and a broader array of credential options compared to my initial approach.

For the Loco implementation, I propose a slight modification. I'll be uploading a revised version shortly and will tag you again for your valuable feedback.

PR #408, opened 2024-02-01
Allow required UUID type in scaffold.

BWStearns:
Allow scaffolding to create a required UUID.


BWStearns:
I just realized the existing UUID field was mandatory by default (unless I'm mistaken) so adding optionality here will require going back and making all existing usages of it into the `uuid!` version.

jondot:
@BWStearns your current fix looks OK, what do you mean?

BWStearns:
Ah I misread the TYPEMAP bit in the gen model [here](https://github.com/loco-rs/loco/blob/3ae219a49bd7552fd8ca6011c7940744a7b27577/src/gen/model.rs#L22-L25). Yeah I guess it should be fine.

Issue #410, opened 2024-02-01
Error: non-exhaustive patterns: `&DatabaseConnection::SqlxMySqlPoolConnection(_)` not covered


Borderliner:
**Description**

Cannot build loco with `sqlx-mysql`, because apparently `pattern `&DatabaseConnection::SqlxMySqlPoolConnection(_)` not covered` in file `src/db.rs` which seems to be true.

**To Reproduce**

Use the following config for development:
`uri: {{ get_env(name="DATABASE_URL", default="mysql://root@127.0.0.1:3306/smm_loco_dev") }}`

**Expected Behavior**

Build successfully

**Environment:**

* I'm on Linux Mint 21.3 which is based on Ubuntu 22.04 LTS so there's a minimum chance of incompatibility.
* Using `rustc 1.75.0 (82e1608df 2023-12-21)`
* Tried Loco v0.2.3 or v0.2.2
* Using Devilbox docker container to access database on 127.0.0.1:3306 which works just fine.

**Additional Context**
```
error[E0004]: non-exhaustive patterns: `&DatabaseConnection::SqlxMySqlPoolConnection(_)` not covered
  --> /home/reza/.cargo/registry/src/index.crates.io-6f17d22bba15001f/loco-rs-0.2.3/src/db.rs:41:11
   |
41 |     match db {
   |           ^^ pattern `&DatabaseConnection::SqlxMySqlPoolConnection(_)` not covered
   |
note: `DatabaseConnection` defined here
  --> /home/reza/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sea-orm-0.12.12/src/database/db_connection.rs:19:1
   |
19 | pub enum DatabaseConnection {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
22 |     SqlxMySqlPoolConnection(crate::SqlxMySqlPoolConnection),
   |     ----------------------- not covered
   = note: the matched value is of type `&DatabaseConnection`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
   |
58 ~         },
59 +         &DatabaseConnection::SqlxMySqlPoolConnection(_) => todo!()
   |

For more information about this error, try `rustc --explain E0004`.
error: could not compile `loco-rs` (lib) due to previous error
```

PR #411, opened 2024-02-01
fix: issue #410 in db.rs

Borderliner:
Fix the issue #410 where one could'nt use mysql for the driver because `DatabaseConnection` match case was not fully covered.


GreenAppleOne:
"When will the new version be released? I am also encountering this issue at the moment, the code generated by the CLI does not support MySQL database."

kaplanelad:
@GreenAppleOne  we're actively developing version 3.0.0 and anticipate completing several planned features by next week

PR #413, opened 2024-02-02
Correct scaffolding example to use correct type.

BWStearns:
Scaffolding uses `string` not text. Rails uses `string` as well so that seems consistent. This PR just corrects the example.


jondot:
Using `text` in the example serves 2 purposes:

1. Within the rails context, text usually generated a textarea (for a blog post)
2. Here in the docs, it serves to show there are more than just 1 type of field, so it is educative (if everything is string a user might thing they should always use `string` for everything)

This is why I believe the example should remain as is. But, is there anything else you noticed that i missed?

BWStearns:
I will double check the behavior later tonight but I believe using the text field in scaffold yielded a broken migration and broken code. my suggestion to use `string` was a result of that failure. I'll reproduce it tonight and look for a better solution.

Issue #409, opened 2024-02-01
`cargo run -- down` doesn't revert a migration


BWStearns:
**Description**

The instruction to run a down migration is [here](https://github.com/loco-rs/loco/blob/eb7b7bdc3b2bfb0d9217655119068a629cc1f17c/starters/saas/migration/README.md?plain=1#L20) but it doesn't work.

The result is

```
warning: `my_app` (lib) generated 11 warnings (run `cargo fix --lib -p my_app` to apply 10 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 0.35s
     Running `target/debug/my_app-cli down`
error: unrecognized subcommand 'down'

Usage: my_app-cli [OPTIONS] <COMMAND>
```

**To Reproduce**

Run `cargo run -- down`

**Expected Behavior**

Revert last migration

**Environment:**

Dev

jondot:
The migrator framework is actually very powerful, see here for the `up` direction (from Loco codebase)

<img width="501" alt="image" src="https://github.com/loco-rs/loco/assets/83390/9a9379fe-32a8-4286-9b04-245ff2da45ba">

We tried to hide the complexity and offer only "up", assuming mostly people fix "up" mistakes in production by applying another migration on top of the bad one -- thus, still go always "up".
Rarely did I use "down" migrations in Rails, so we're trying to delay their use as much as possible.

If you'd like to submit a PR that implements the "down" part we'd be happy to accept (it's mostly just copy-paste code and issuing calls to the migrator framework)

BWStearns:
I'll get a PR for that in. I will never in my life get a generation command right on the first try so it's important for me to be able to run the down command for convenience 😆

BWStearns:
Started it [here](https://github.com/loco-rs/loco/pull/414)

BWStearns:
Finished up the down command: https://github.com/loco-rs/loco/pull/414

kaplanelad:
this issue was merge and released

PR #414, opened 2024-02-02
Add a down migration command

BWStearns:
I think this should add the down migration command. I still want to add docs and at least test it locally.

### DIFF
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -169,15 +169,19 @@ This enforces _everything-as-code_, _reproducibility_ and _atomicity_, where no 
 
 If you realize that you made a mistake, you can always undo the migration. This will undo the changes made by the migration (assuming that you added the appropriate code for `down` in the migration).
 
+<!-- <snip id="migrate-down-command" inject_from="yaml" template="sh"> -->
+```sh
+cargo loco db down
 ```
-$ cargo loco db down
-```
+<!-- </snip> -->
 
 The `down` command on its own will rollback only the last migration. If you want to rollback multiple migrations, you can specify the number of migrations to rollback.
 
+<!-- <snip id="migrate-down-n-command" inject_from="yaml" template="sh"> -->
+```sh
+cargo loco db down 2
 ```
-$ cargo loco db down 2
-```
+<!-- </snip> -->
 
 ### Verbs, singular and plural
 
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -165,6 +165,19 @@ Loco is a migration-first framework, similar to Rails. Which means that when you
 
 This enforces _everything-as-code_, _reproducibility_ and _atomicity_, where no knowledge of the schema goes missing. 
 
+### Down Migrations
+
+If you realize that you made a mistake, you can always undo the migration. This will undo the changes made by the migration (assuming that you added the appropriate code for `down` in the migration).
+
+```
+$ cargo loco db down
+```
+
+The `down` command on its own will rollback only the last migration. If you want to rollback multiple migrations, you can specify the number of migrations to rollback.
+
+```
+$ cargo loco db down 2
+```
 
 ### Verbs, singular and plural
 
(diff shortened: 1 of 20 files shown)


jondot:
Looks good, I would recommend to verify:

1. Down should be probably 1 step down. when supplying `None` we need to make sure it doesn't go all the way back for all migrations
2. Tie the command back to the CLI (follow what `up` does)

kaplanelad:
related issue: https://github.com/loco-rs/loco/issues/409

BWStearns:
I finally got around to knocking this out. I tested it and it works. Running `cargo loco db down` will undo one migration, and specifying a number `cargo loco db down n` will undo n migrations.

BWStearns:
Just tagging @kaplanelad since this PR might have justifiably been presumed dead.

src/cli.rs
```diff
@@ -203,6 +203,11 @@ enum DbCommands {
     Create,
     /// Migrate schema (up)
     Migrate,
+    /// Run one down migration, or add a number to run multiple down migrations (i.e. `down 2`)
+    Down {
+        /// The number of migrations to rollback
+        steps: Option<u32>,
```
kaplanelad, line 209: Please remove the option<u32> and add clippy default as 1.  
this will be clear when users use --help and see what is the default value

src/boot.rs
```diff
@@ -126,6 +126,8 @@ pub async fn run_task<H: Hooks>(
 pub enum RunDbCommand {
     /// Apply pending migrations.
     Migrate,
+    /// Run one or more down migrations.
+    Down(Option<u32>),
```
kaplanelad, line 130: Following the CLI default value you can set the migration number will be not an option

docs-site/content/docs/the-app/models.md
```diff
@@ -165,6 +165,19 @@ Loco is a migration-first framework, similar to Rails. Which means that when you
 
 This enforces _everything-as-code_, _reproducibility_ and _atomicity_, where no knowledge of the schema goes missing. 
 
+### Down Migrations
+
+If you realize that you made a mistake, you can always undo the migration. This will undo the changes made by the migration (assuming that you added the appropriate code for `down` in the migration).
+
+```
+$ cargo loco db down
+```
```
kaplanelad, line 174: Please move this (also the second command) command to take it from [snipdoc. yml](https://github.com/loco-rs/loco/blob/master/snipdoc.yml).
This is a tool that we built to improve our documentation. See how to use this tool [here](https://github.com/kaplanelad/snipdoc)

src/db.rs
```diff
@@ -182,6 +182,18 @@ pub async fn migrate<M: MigratorTrait>(db: &DatabaseConnection) -> Result<(), se
...
 }
 
+/// Revert migrations to the database using the provided migrator.
+///
+/// # Errors
+///
+/// Returns a [`sea_orm::DbErr`] if an error occurs during run migration up.
+pub async fn down<M: MigratorTrait>(
+    db: &DatabaseConnection,
+    steps: Option<u32>,
```
kaplanelad, line 192: following the steps changes, change it as well

kaplanelad (commented):
Thank you, @BWStearns, for this PR. I've left a few comments—please take a look when you have a chance.

src/cli.rs
BWStearns: Is there an example of clippy default?

docs-site/content/docs/the-app/models.md
BWStearns: I took a crack at it but it also went after some other code snippets. What should I do there?

BWStearns:
Just a heads up, locally `snipdoc run` broke a few other docs on my machine. Is there some extra setup I need to do?

kaplanelad:
@BWStearns, Can you share with me what is broken?

src/cli.rs
kaplanelad: see example https://github.com/loco-rs/loco/blob/master/src/cli.rs#L145

BWStearns:
Pretty sure this is good to merge since the CI snipdoc is happy. Just wanted to call out the disparity between CI snipdoc and a local run of it.

kaplanelad (approved):

PR #422, opened 2024-02-06
feat: update shuttle deployment template to 0.38

### LINKED ISSUE #395: shuttle deployment: update connection set up code
Shuttle have released a new way to connect, which is preferrable:

https://github.com/shuttle-hq/shuttle-docs/pull/249/files#diff-d4300a8ea6183af183af182922cb2936c879e8baf169117e6460cb2a93e8a50bR75

we need to take that string, set an env variable we require, and then let the app continue booting.

jonaro00:
Closes #395 

- Bumps shuttle versions to 0.38.0, use same version for all shuttle crates.
- Adds a shutte postgres connection if a `with-db` template is used.
- Fixed the environment parsing, and uses environment based on shuttle metadata.

I tested generating and running locally for the lightweight and the saas template.
TODO: Confirm deployment on Shuttle works. (the env + config etc)

### DIFF
--- a/src/gen/templates/deployment_shuttle.t
+++ b/src/gen/templates/deployment_shuttle.t
@@ -16,19 +16,30 @@ injections:
     path = "src/bin/shuttle.rs"
 - into: Cargo.toml
   after: \[dependencies\]
-  content: "shuttle-runtime = { version = \"{{shuttle_runtime_version}}\", default-features = false }"
-- into: Cargo.toml
-  after: \[dependencies\]
-  content: "shuttle-axum = { version = \"{{shuttle_axum_version}}\", default-features = false, features = [\"axum-0-7\",] }"
+  content: |
+    shuttle-axum = "{{shuttle_runtime_version}}"
+    shuttle-metadata = "{{shuttle_runtime_version}}"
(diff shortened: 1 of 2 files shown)


src/gen/templates/deployment_shuttle.t
```diff
@@ -16,19 +16,30 @@ injections:
...
-    let boot_result = create_app::<App>(StartMode::ServerOnly, &environment)
+async fn main(
+  {% if with_db %}#[shuttle_shared_db::Postgres] conn_str: String,{% endif %}
+  #[shuttle_metadata::ShuttleMetadata] meta: shuttle_metadata::Metadata,
+) -> shuttle_axum::ShuttleAxum {
+    {% if with_db %}std::env::set_var("DATABASE_URL", conn_str);{% endif %}
+    let environment = match meta.env {
+        shuttle_metadata::Environment::Local => Environment::Development,
+        shuttle_metadata::Environment::Deployment => Environment::Production,
+    };
```
jonaro00, line 41: Env was always "development" in the context of Shuttle before. Is this the intended way to use the env or does it complicate the config required for deploying on Shuttle?

kaplanelad (approved):

PR #447, opened 2024-02-16
Update Dockerfile

hericah:
sea-orm-cli complaint because rust version on the rust:0-1 is less than minimum requirement


kaplanelad:
This is already fixed by @fan-tastic-z [PR](https://github.com/loco-rs/loco/pull/458).
Thanks @hericah

PR #451, opened 2024-02-17
Add adminer to devcontainer restapi starter

hericah:


kaplanelad:
Thanks, @hericah, for this PR. 
We are prepared to take a side of external tools.
When creating the starters the devcontainer is int your responsibility and you can add it in each project that you starting

PR #456, opened 2024-02-22
Oauth feature - Authorization Code Grant

yinho999:
Completed Documentation, example demo code, library, and Integration/E2E testing for Authorization Code Grant.


jondot:
Great addition!
I will take some time to study the code and give some feedback ❤️

kaplanelad:
@jondot @yinho999 
We need to consider to moving it to a separated crate, with the option to override the configuration.

something like that: https://crates.io/crates/rhai-loco

marcson909:
I've recently been looking into how I could integrate clerk-rs into Loco (the crate only has actix web middleware). This would be a great addition to make it easy to handle the OAuth myself and I don't mind going that route at all but to better my own understanding i'm still curious how we enable both implementations

yinho999:
> @jondot @yinho999 
> We need to consider to moving it to a separated crate, with the option to override the configuration.
> 
> something like that: https://crates.io/crates/rhai-loco

I think create a general security crate for OAuth2, MFA or TOTP and more. Users can just toggle the feature flags for each strategy if they needed.

jondot:
I have reviewed the PR, amazing work @yinho999 a few questions:

1. Could this support providers like Google, Facebook, and others in parallel? (i.e. login with "Google | Facebook | X") i believe there has to be some kind of _provider ID_ concept and a way to configure multiple (like node.js Passport)
2. Can it also work in parallel with the regular JWT auth?

A few notes, I generally agree with @kaplanelad, I would be happy to see it as a "plug in", so try to create an `initializer` that wires this functionality that would be the best. Once you have it as an initializer, it can be consumed as a crate for example.

yinho999:
> I have reviewed the PR, amazing work @yinho999 a few questions:
> 
> 1. Could this support providers like Google, Facebook, and others in parallel? (i.e. login with "Google | Facebook | X") i believe there has to be some kind of _provider ID_ concept and a way to configure multiple (like node.js Passport)
> 2. Can it also work in parallel with the regular JWT auth?
> 
> A few notes, I generally agree with @kaplanelad, I would be happy to see it as a "plug in", so try to create an `initializer` that wires this functionality that would be the best. Once you have it as an initializer, it can be consumed as a crate for example.

1. Yes, the current code I am running can be ran with Google credentials(OAuth2 Authorization Code). Since the OAuth2 is a standard, we should able to use any provider with Authorization Code Grant supported. Futhermore, I added provider name in the */yaml, which store as a key in the OAuth2Store BTreeMap. User can get the specific OAuth2 instance from the OAuth2 store BTreeMap.(I might realize this approach might cause conflict between different grant but same provider, I will change the naming of this.)

2. Yes, the login logic is completely different than jwt token, which means we need different endpoints to jwt token which works parallel to JWT.


I will start working on converting this code into a plugin/initializer few days later. I sick these few days and I will start after I get better.👍

yinho999:
starting to work on it right now

yinho999:
@jondot @kaplanelad Hello there, after exploring the initializer pattern, I'm somewhat unclear about the appropriate location for the configuration file. Currently, the OAuth2 configuration structure is placed within the `src/configuration.rs` file, and the variables are in the `config/[..].yaml` file. 

However, if the OAuth2 functionality is to be separated into a different library, I'm uncertain where the configuration file and variables should be located. One solution might be to isolate the configuration into a separate file and access it through a PathBuf, although I'm not convinced that's the best approach since this approach requires the user to copy and paste the config structure by hand.

kaplanelad:
Hey @yinho999, 
You can move the configuration via [settings config](https://loco.rs/docs/getting-started/config/#settings).

PR #463, opened 2024-02-26
reshape

jondot:


kaplanelad:
@jondot I have separated the query from the exec. change the demo app to using the new strucher. 

Still needs:
- [ ] adding order by in exec 
- [ ] move the Entity from pagination into exec
   from
   ```rs
   query::exec(&db).page(2).page_size(100).paginate::<test_db::Entity>()
   ```
   to
      ```rs
   query::exec::<test_db::Entity>(&db).page(2).page_size(100).paginate()
   ```

PR #481, opened 2024-03-07
Allow trailing slashes in URIs by adding the NormalizePathLayer

spencewenski:
Problem
-------
Some clients may append a trailing slash to URIs, either intentionally or accidentally. We should support both routes that do and don't have trailing slashes for maximum compatibility.

Solution
--------
Normally, when we define a route, it will not have a trailing slash. We can add support for trailing slashes by adding the `NormalizePathLayer`, which will automatically remove the trailing slash from the requested URI, allowing it to map to the route we created without a trailing slash.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ tower-http = { version = "0.5.0", features = [
   "fs",
   "set-header",
   "compression-full",
+  "normalize-path",
 ] }
 byte-unit = "4.0.19"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,7 +132,6 @@ tower-http = { version = "0.5.0", features = [
   "fs",
   "set-header",
   "compression-full",
-  "normalize-path"
 ] }
 
 [dependencies.sea-orm-migration]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,17 +86,7 @@ tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 # cli/migrations
 duct = "0.13.6"
 
-tower-http = { version = "0.5.0", features = [
-  "trace",
-  "catch-panic",
-  "timeout",
-  "add-extension",
-  "cors",
-  "fs",
-  "set-header",
-  "compression-full",
-  "normalize-path",
-] }
+tower-http = { workspace = true }
 byte-unit = "4.0.19"
 
 argon2 = { version = "0.5.2", features = ["std"] }
@@ -105,7 +95,7 @@ jsonwebtoken = { version = "9.1.0", optional = true }
 bcrypt = { version = "0.15.0", optional = true }
 validator = { version = "0.16.1", features = ["derive"] }
 futures-util = "0.3"
-tower = "0.4"
+tower = { workspace = true }
 hyper = "1.1"
 mime = "0.3"
 bytes = "1.1"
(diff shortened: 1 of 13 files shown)


kaplanelad:
@spencewenski can we move this logic to [initializer](https://github.com/loco-rs/loco/tree/master/loco-extras) and add it as a default in our starter. this is allows to user to turn it off if they don't want it

spencewenski:
Hi @kaplanelad, thanks for the suggestion! I got a prototype of the initializer approach working. Below is a snippet of the code. My only concern is I'm not sure if there's any implications to routing when using this approach. I'll read some docs and play around with this some more and if everything seems okay I'll update the PR with the initializer approach.

Code snippet:
```rust
async fn after_routes(&self, router: Router, _ctx: &AppContext) -> Result<Router> {
    let router = NormalizePathLayer::trim_trailing_slash().layer(router);
    let router = Router::new().nest_service("", router);
    Ok(router)
}
```

kaplanelad:
You can follow loco-extras [here](https://github.com/loco-rs/loco/blob/master/loco-extras/src/initializers/mod.rs).
The initializer gets `AxumRouter`. 

[Here ](https://github.com/loco-rs/loco/blob/master/examples/demo/src/app.rs#L53) you can see use initializer with `prometheus`. 

i'm suggestion to do the same.

spencewenski:
Yup, I understand. I updated the PR. My concerns were related to the behavior of `nest_service` -- specifically that the [nested routes have the prefix stripped](https://docs.rs/axum/latest/axum/struct.Router.html#how-the-uri-changes). However, since we're nesting at the root route, this doesn't appear to be a problem.

Cargo.toml
```diff
@@ -131,6 +122,18 @@ object_store = { version = "0.9.0", default-features = false }
...
+tower-http = { version = "0.5.0", features = [
+  "trace",
+  "catch-panic",
+  "timeout",
+  "add-extension",
+  "cors",
+  "fs",
+  "set-header",
+  "compression-full",
+  "normalize-path"
```
kaplanelad, line 135: This is not a requirement from loco framework, this means that the `normalize-path` will includded for all the users even if they not using your initializer. can you please move it to the loco-extras cargo file?

examples/demo/src/app.rs
```diff
@@ -46,6 +47,7 @@ impl Hooks for App {
             Box::new(initializers::axum_session::AxumSessionInitializer),
             Box::new(initializers::view_engine::ViewEngineInitializer),
             Box::new(initializers::hello_view_engine::HelloViewEngineInitializer),
+            Box::new(loco_extras::initializers::normalize_path::NormalizePathInitializer),
```
kaplanelad, line 50: Can you please add a test case in `https://github.com/loco-rs/loco/tree/master/examples/demo/tests/requests`?

kaplanelad (changes requested):
Thank you, @spencewenski! This is exactly what I was looking for.
Could you please take a look to review two small notes?

examples/demo/src/app.rs
spencewenski: Added tests using the `ping` endpoint.

Cargo.toml
spencewenski: Done. Moved to the `loco-extras` cargo file.

kaplanelad (approved):

spencewenski:
Hey, @kaplanelad , thanks approving the PR! Are you able to merge it for me? I don’t have push permissions.

jondot:
@kaplanelad I’m wondering is this useful as a default inside loco for all routes?

jondot:
I mean why not include it not as an initializer but as an integrated later in all routers instead of our custom code for handing ‘//‘?

PR #482, opened 2024-03-07
Fix the doc test for `src/controller/mod.rs`

spencewenski:
Problem
-------
The sample rust code in `src/controller/mod.rs` does not compile with a fresh clone of `loco` because it is using some methods and imports that are only available when the `channels` feature is enabled.

Solution
--------
Update the sample code to guard the offending blocks of code with a `cfg` guard.


spencewenski:
Nvm, I was using just `cargo test`, but I see CI is using `cargo test --all-features --all`, which allows the doc tests to pass.

PR #484, opened 2024-03-07
wip: respond_to and response overhaul

jondot:
- Moving to `Result<Response>` creates more fluent code and moving between return type without requiring the user for specifying types.
- Add `respond_to` responder based on incoming content type


jondot:
@kaplanelad I think this is a correct direction.

take a look when you can, see if there's anything i've missed

src/gen/templates/controller.t
```diff
@@ -18,7 +18,7 @@ pub async fn echo(req_body: String) -> String {
     req_body
 }
 
-pub async fn hello(State(_ctx): State<AppContext>) -> Result<String> {
+pub async fn hello(State(_ctx): State<AppContext>) -> Result<Response> {
     // do something with context (database, etc)
     format::text("hello")
```
assapir, line 23: missing `into_response()` here, no?

kaplanelad:
@jondot we also need to support the content type when returns an errors.
i'll create a PR based this pr for an example

PR #486, opened 2024-03-07
Use Into<String> in function signature to allow String to be passed in directly

PolpOnline:


PolpOnline:
@jondot should be fixed now

PR #489, opened 2024-03-08
Warn when a user creates a new project inside a exisiting git repo

### LINKED ISSUE #487: loco-cli should warn when trying to create an app inside another git repository
## Feature Request

Every developer did it at least once in their life - starting a new project (`npm init`/`loco new`) inside an already existing project, then realizing it too late and starting moving things.

I would like the loco-cli to warn the user when running the command if there is a `.git` folder up in the directory tree and ask them explicitly to allow that.

1st alternative: As of today, ignore that, which I obviously don't think it's a good idea.
2nd alternative: Don't allow that at all. I think this alternative is too strict to the other side when sometimes the user does want to do that.
  - This can be done by allowing an explicit flag to the command line, but because it's already interactive, why not keep it interactive?

As a side note, I would love to help develop this feature request with appropriate mentoring.

assapir:
Closes #487.
I tested it manually, I couldn't find any automated tests for the CLI

### DIFF
--- a/loco-cli/src/git.rs
+++ b/loco-cli/src/git.rs
@@ -144,6 +144,23 @@ fn git_exists() -> bool {
     }
 }
 
+pub fn is_a_git_repo(destination_path: &PathBuf) -> eyre::Result<bool> {
+    let destination_path = destination_path.canonicalize()?;
+    match Command::new("git")
+        .arg("-C")
+        .arg(destination_path)
+        .arg("rev-parse")
+        .arg("--is-inside-work-tree")
+        .output()
+    {
+        Ok(output) => Ok(output.status.success()),
+        Err(err) => {
+            tracing::debug!(error = err.to_string(), "git not found");
+            Ok(false)
+        }
+    }
+}
+
 #[cfg(feature = "git2")]
 fn clone_repo_with_git2(temp_clone_dir: &Path) -> eyre::Result<()> {
     let mut fetch_options = git2::FetchOptions::new();
--- a/loco-cli/src/git.rs
+++ b/loco-cli/src/git.rs
@@ -153,7 +153,17 @@ pub fn is_a_git_repo(destination_path: &PathBuf) -> eyre::Result<bool> {
         .arg("--is-inside-work-tree")
         .output()
     {
-        Ok(output) => Ok(output.status.success()),
+        Ok(output) => {
+            if output.status.success() {
+                Ok(true)
+            } else {
+                tracing::error!(
+                    error = tracing::field::debug(output),
+                    "git command returned an error"
+                );
+                Ok(false)
+            }
+        }
         Err(err) => {
             tracing::debug!(error = err.to_string(), "git not found");
             Ok(false)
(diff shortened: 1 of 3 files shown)


loco-cli/src/prompt.rs
```diff
@@ -78,6 +78,36 @@ pub fn template_selection(
...
+
+    let res = requestty::prompt_one(question)?;
+    let answer = res
+        .as_bool()
+        .ok_or_else(|| eyre::eyre!("allow_git_repo is empty"))?;
+
+    if answer {
+        Ok(())
+    } else {
+        Err(eyre::eyre!("user choose not to continue"))
```
jondot, line 107: Maybe a more personal "aborted: you've chosen to quit."

jondot (approved):

loco-cli/src/git.rs
```diff
@@ -144,6 +144,23 @@ fn git_exists() -> bool {
...
+pub fn is_a_git_repo(destination_path: &PathBuf) -> eyre::Result<bool> {
+    let destination_path = destination_path.canonicalize()?;
+    match Command::new("git")
+        .arg("-C")
+        .arg(destination_path)
+        .arg("rev-parse")
+        .arg("--is-inside-work-tree")
+        .output()
+    {
+        Ok(output) => Ok(output.status.success()),
```
kaplanelad, line 156: let logged the error if the command returned an error
```suggestion
     Ok(output) => {
            if output.status.success() {
                Ok(true)
            } else {
                tracing::error!(
                    error = tracing::field::debug(output),
                    "git command returned an error"
                );
                Ok(false)
            }
        }
```

loco-cli/src/prompt.rs
```diff
@@ -78,6 +78,36 @@ pub fn template_selection(
...
+/// If the `ALLOW_IN_GIT_REPO` environment variable is set, this test will be skipped.
+/// If the environment variable is not set, the function will warn the user if they are inside a git
+/// and will let them cancel the operation or continue.
+///
+/// # Errors
+/// when could not prompt the question, or when the user choose not to continue.
+pub fn warn_if_in_git_repo() -> eyre::Result<()> {
+    if let Ok(_) = env::var("ALLOW_IN_GIT_REPO") {
+        return Ok(());
+    }
```
kaplanelad, line 92:
```suggestion
    if env::var("ALLOW_IN_GIT_REPO").is_ok() {
        return Ok(());
    }
```

loco-cli/src/git.rs
kaplanelad: Id the command not succeeded let print the strerr to log.

kaplanelad (commented):
Hey @assapir good to see you here :)!
See small comments

assapir:
> Hey @assapir good to see you here :)! See small comments

Thanks @kaplanelad, I said I would open PRs, so I did :)
Fixed regarding your comments, please take a look again.

kaplanelad (approved):

kaplanelad:
@thanks @assapir you are man of word.
i'll publish the cli later on today

assapir:
> @thanks @assapir you are man of word. i'll publish the cli later on today

Do you want me to create a GH action for automating that? if so, just create a secret with the creds needed and I will add that :)

PR #491, opened 2024-03-08
Put the loco-cli as workspace member

assapir:
It makes things easier for me when working on loco-cli: I don't have to open it as a separate project, and can work on it when opening the root dir.
I can understand if you decide not to accept this  PR


jondot:
Actually we want at some stage to put it in its own repo, in order to strengthen the fact that it's a "dumb generator", making sure no Loco-specific logic gets mixed into it, or other way around.

assapir:
Cool, that makes sense. I will close this one, but please move it to it's own repo then :)

Issue #490, opened 2024-03-08
Rust analyzer shows errors with loco-rs


assapir:
**Description**

When trying to open the `loco-rs` in VSCode with rust-analyzer extension, this extension fails to do its job, with the following errors:
```
2024-03-08T11:49:25.833538Z ERROR project_model::workspace: cyclic deps: loco_rs(Idx::<CrateData>(197)) -> loco_rs(Idx::<CrateData>(197)), alternative path: loco_rs(Idx::<CrateData>(197))
2024-03-08T11:49:26.301369Z ERROR project_model::workspace: cyclic deps: loco_rs(Idx::<CrateData>(197)) -> loco_rs(Idx::<CrateData>(197)), alternative path: loco_rs(Idx::<CrateData>(197))
2024-03-08T11:49:27.030212Z ERROR project_model::workspace: cyclic deps: loco_rs(Idx::<CrateData>(197)) -> loco_rs(Idx::<CrateData>(197)), alternative path: loco_rs(Idx::<CrateData>(197))
```

**To Reproduce**

Open the project in VSCode with rust-analyzer installed on a Linux (x64)

**Expected Behavior**

rust-analyzer should work and give all it's expected functionality, like "go to definitions", etc.

**Environment:**

rust-analyzer version details (from it's logs):
```
INFO [08/03/2024, 13:49:25]: /home/assaf/.vscode/extensions/rust-lang.rust-analyzer-0.3.1868-linux-x64/server/rust-analyzer --version: {
  status: 0,
  signal: null,
  output: [ null, 'rust-analyzer 0.3.1868-standalone\n', '' ],
  pid: 984450,
  stdout: 'rust-analyzer 0.3.1868-standalone\n',
  stderr: ''
}
```

**Additional Context**
Probably related to https://github.com/rust-lang/rust-analyzer/issues/14167

I can help debug on my machine if needed. This doesn't happen if one is only opening an internal crate like the `loco-cli`

assapir:
> Never had that one, I'm using pre-release usually, which only failed me once in 3-4 years.
> 
> rust-analyzer 0.4.1870-standalone (767d5d3ea 2024-03-05)
> 
> @kaplanelad are you using pre-release or release?

I tried to switch to pre-release, still happens:
```
INFO [08/03/2024, 14:25:48]: /home/assaf/.vscode/extensions/rust-lang.rust-analyzer-0.4.1873-linux-x64/server/rust-analyzer --version: {
  status: 0,
[...]

PR #492, opened 2024-03-08
uncycle rust-analyzer

jondot:
see #490 

bringing in `tests_cfg` the general purpose test data structures that are used in the new query DSL, to the default scope.

considered but didn't take action:
1. put `tests_cfg` under a `cfg(test)` which seems to be aligned with the usage, but then, users of the frameworks can't enjoy it
2. put `tests_cfg` in a `loco-test` crate, but didn't do it because it's too small. might be worth while to do it anyway, and then include this crate as a dev-dep.

@kaplanelad LMK what you think


kaplanelad:
I think it better to move it to internal crate for testing. this is allows as to improve our tests in `Migration` and `Hook` test app. 

i'll open a pr for that and then close this PR

PR #496, opened 2024-03-09
Response body content

yinho999:
Corrected the handling of most API error responses to ignore the error content and only return the standardized body with `"error": "Bad Request"`. Additionally, altered the `Error::message` to return plain text instead of a JSON type. Implemented testing for API errors.

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -276,6 +276,8 @@ fn json_error_response(status_code: StatusCode, detail: ErrorDetail) -> Response
 
 #[cfg(test)]
 mod tests {
+    use std::env;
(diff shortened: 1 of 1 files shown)


schungx:
Well, a suggestion: look at the `Accept-Encoding` header... if it is `application/json`, then return JSON, otherwise plain text...

kaplanelad:
@yinho999,
@jondot already working on this, see this PR: #484

yinho999:
> @yinho999, @jondot already working on this, see this PR: #484

Sorry for not check existed ticket. Thanks for reminding me.👍️

Issue #501, opened 2024-03-17
docs: typo and formatting mistakes


AndreiOrmanji:
**Description**

I've read everything from [docs](https://loco.rs/docs/) and I have found some small mistakes:

- https://loco.rs/docs/features/authentication/#get-current-user
![image](https://github.com/loco-rs/loco/assets/48985454/4591315d-82b1-42b9-a163-8e9439a1dee1)

- https://loco.rs/docs/getting-started/deployment/
![image](https://github.com/loco-rs/loco/assets/48985454/f15a1fbd-41e0-4400-a726-c66c52ac2f08)

- https://loco.rs/docs/help/faq/
![image](https://github.com/loco-rs/loco/assets/48985454/44016646-ab2a-41d9-8c9b-5f1eb18469e2)

- https://loco.rs/docs/testing/overview/#clean-up-database-for-each-test
![image](https://github.com/loco-rs/loco/assets/48985454/dac511f8-ec56-4d4b-9083-1165784f3044)

- https://loco.rs/docs/the-app/mailers/#mailer-configuration
![image](https://github.com/loco-rs/loco/assets/48985454/97941764-596a-4e6f-b6d1-6ca824047d5f)

PR #502, opened 2024-03-17
docs: typo and formatting mistakes fixes

AndreiOrmanji:
Closes https://github.com/loco-rs/loco/issues/501

### DIFF
--- a/docs-site/content/docs/help/faq.md
+++ b/docs-site/content/docs/help/faq.md
@@ -16,7 +16,8 @@ flair =[]
 
 <details>
 <summary>How can I automatically reload code?</summary>
-Try (cargo watch)[https://crates.io/crates/cargo-watch]:
+
+Try [cargo watch](https://crates.io/crates/cargo-watch):
 
 ```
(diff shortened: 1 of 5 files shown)


docs-site/content/docs/help/faq.md
```diff
@@ -16,7 +16,7 @@ flair =[]
 
 <details>
 <summary>How can I automatically reload code?</summary>
-Try [cargo watch](https://crates.io/crates/cargo-watch):
+Try (cargo watch)[https://crates.io/crates/cargo-watch]:
```
jondot, line 19: looks like this should be the opposite brackets?

AndreiOrmanji: yes, you are right. I've changed it because it was not rendering, but issue was different: [the rule of adding newlines before and after blocks](https://docs.gitlab.com/ee/user/markdown.html#details-and-summary:~:text=Remember%20to%20leave%20a%20blank%20line%20before%20and%20after%20any%20Markdown%20sections) was not respected. I will fix it with next commit.

jondot:
Thanks!

PR #507, opened 2024-03-20
feat: update query dsl condition

didy-kpn:
I want to use "is_in/is_not_in" to condition when paginating, so I add it.
This is my first contribution, so I would appreciate any feedback if I did something wrong.

```rust
let condition = model::query::dsl::condition()
    .is_in(notes::Column::TagId, [1]); // I want to use "is_in"

let paginated_notes = model::query::exec::paginate(
    &ctx.db,
    notes::Entity::find(),
    Some(condition.build()),
    &pagination_query,
)
.await?;
```


kaplanelad:
Thanks @didy-kpn

Issue #520, opened 2024-04-04
Replace Redis with Valkey (or other alternatives)


simonsan:
Due to recent license changes, it may hinder adoption of loco, that it pulls in Redis in its SaaS template, maybe a change to Valkey could be useful?

See https://www.theregister.com/2024/04/03/open_source_redis_alternative/

jondot:
I believe most of this would be docs changes, right?

Muxoid:
I created a pull request if this is the direction that the project wants to take, [the pull request](https://github.com/loco-rs/loco/pull/523).

PR #523, opened 2024-04-06
Changed all references from redis to valkey

Muxoid:
This is related to issue #520 at this moment the docker container is in unstable branch. I tested the examples/demo and all tests succeeded I tried testing the root directory but it failed in both my branch and the loco-rs main branch.

I saw the issue and saw the changes to the Redis licensing, I am new to loco development but thought I could help out with this change if this is something that would interest you, and aligns with the direction of the project.


jondot:
Thanks for the thorough refactor!
A few questions might pose a dilemma:

* Is the protocol itself called "Redis protocol"?
* If we think of a client, is it a Redis client?

if so, I believe we should keep most of the code-related Redis references such as the `redis:` section in config and `redis` field names, and then instruct people to install `valkey` indicating our strong support of open projects, but also make a note that if they want, they can use the regular Redis. I imagine some people would still buy Redis services with RedisLabs (the entity that changed the license).

Muxoid:
That makes sense, in the end the only thing that would need to change is the Docker file to decouple from RedisLabs.

You can reject it if you believe this is too much of a change.

jondot:
@Muxoid I'm happy if you could make a PR with just the Docker recommendation and some guidance in the docs, which may be less intrusive. Would be highly appreciated!

Muxoid:
I've updated the documentation and docker-compose.yml files to include Valkey as an option, without altering the default configurations.

PR #525, opened 2024-04-09
Strip trailing whitespace from email subject templates

ZymoticB:
I ran into a surprising issue with the SaaS starter. I've been doing some game development recently, which has led to me developing on Windows. I ran into some issues with line endings in CI, which I wasn't too surprised about, but when I started digging I was only finding issues that had already been resolved!

I finally tracked my issue down to how https://github.com/lettre/lettre formats emails. `\n` and `\r` are (10 and 13) are both skiiped here https://github.com/lettre/lettre/blob/22dca340a7bfc338f9925b0d7c4ecf915daa7e4f/src/message/header/mod.rs#L405. lettre is certainly doing the right thing here utf-8 encoding these characters, but it's pretty inconvenient since the encoding ends up in the snapshots!

I think it's probably a reasonable tradeoff since I don't actually think users want to be unknowingly sending emails with a newline in the subject just because their editor automatically adds a newline to their template file!

I made change to the email_sender.rs tests to demonstrate what this looks like in a snapshot in practice, and then in a second commit made a small change to template engine to strip trailing whitespace of Subject templates hopefully avoiding surprises for users in the future.


ZymoticB:
I think I may have messed up the snapshots trying to make sure I had a demonstration. I'll take a look later today.

ZymoticB:
I updated the snapshots in `examples/demo`. I didn't realize something was using the in-tree version of loco. I had checked the starters and they use a crate. The starters will need their snapshots updated if this lands and the crate version there is bumped.

I fixed an issue with the cli tests in the demo on windows (there is an exe suffix in the help output). I still had failing tests due to I think terminal color code mismatches? I don't think it will cause issues with CI so I haven't investigated it further, but here is an example failure in case something stands out to someone who is more familiar

```
Testing tests\cmd\cli.trycmd:127 ... failed
Exit: success

---- expected: stdout
++++ actual:   stdout
   1      - [0m  INFO app: loco_rs::config: loading environment from selected_path="config/teste2e.yaml" environment=teste2e
   2      - [0m  WARN app: loco_rs::boot: status: environment=teste2e
   3      - [0m  INFO app: sea_orm_migration::migrator: Checking migration status environment=teste2e
   4      - [0m  INFO app: sea_orm_migration::migrator: Migration 'm20220101_000001_users'... Applied environment=teste2e
   5      - [0m  INFO app: sea_orm_migration::migrator: Migration 'm20231103_114510_notes'... Applied environment=teste2e
        1 + 2024-04-09T14:32:15.771364Z  INFO app: loco_rs::config: loading environment from selected_path="config//teste2e.yaml" environment=teste2e
        2 + 2024-04-09T14:32:15.897624Z  WARN app: loco_rs::boot: status: environment=teste2e
        3 + 2024-04-09T14:32:15.901561Z  INFO app: sea_orm_migration::migrator: Checking migration status environment=teste2e
        4 + 2024-04-09T14:32:15.911769Z  INFO app: sea_orm_migration::migrator: Migration 'm20220101_000001_users'... Applied environment=teste2e
        5 + 2024-04-09T14:32:15.911851Z  INFO app: sea_orm_migration::migrator: Migration 'm20231103_114510_notes'... Applied environment=teste2e
```

PR #533, opened 2024-04-11
Fix wrong path in tour commands

chmelevskij:
When following the tour `posts` seem to be created without the `api` prefix.

Screenshot 2024-04-11 at 15 59 11

This seems to be throughout the whole guide so not sure if it's the docs or the scaffolding commands.

Also missing `from` in MVC description

### DIFF
--- a/docs-site/content/docs/getting-started/guide.md
+++ b/docs-site/content/docs/getting-started/guide.md
@@ -150,7 +150,7 @@ listening on port 3000
 And now, let's see that it's alive:
 
 ```sh
-$ curl localhost:3000/api/_ping
+$ curl localhost:3000/_ping
 {"ok":true}
 ```
 
@@ -159,7 +159,7 @@ The built in `_ping` route will tell your load balancer everything is up.
 Let's see that all services that are required are up:
 
 ```sh
-$ curl localhost:3000/api/_health
+$ curl localhost:3000/_health
 {"ok":true}
 ```
 
--- a/docs-site/content/docs/getting-started/guide.md
+++ b/docs-site/content/docs/getting-started/guide.md
@@ -504,9 +504,9 @@ Edit `src/controllers/articles.rs`:
 #![allow(clippy::unused_async)]
 use loco_rs::prelude::*;
 
-use crate::models::_entities::articles;
+use crate::models::_entities::articles::{self, Model};
 
-pub async fn list(State(ctx): State<AppContext>) -> Result<Response> {
+pub async fn list(State(ctx): State<AppContext>) -> Result<Json<Vec<Model>>> {
     let res = articles::Entity::find().all(&ctx.db).await?;
     format::json(res)
 }
@@ -525,7 +525,7 @@ $ cargo loco start
 And make a request:
 
 ```sh
-$ curl localhost:3000/api/articles
+$ curl localhost:3000/articles
 [{"created_at":"...","updated_at":"...","id":1,"title":"how to build apps in 3 steps","content":"use Loco: https://loco.rs"}]
 ```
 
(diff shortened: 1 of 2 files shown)


docs-site/content/docs/getting-started/guide.md
```diff
@@ -504,9 +504,9 @@ Edit `src/controllers/articles.rs`:
 #![allow(clippy::unused_async)]
 use loco_rs::prelude::*;
 
-use crate::models::_entities::articles;
+use crate::models::_entities::articles::{self, Model};
 
-pub async fn list(State(ctx): State<AppContext>) -> Result<Response> {
```
chmelevskij, line 509: Not sure if this correct or just missing `use axum::response::Response;` at the top?

chmelevskij:
Ok. so there seems more than just that issue. Guide is looking slightly outdated in few cases regarding paths, types etc.

Would like to confirm that's before I put any more time into this

cozyGalvinism:
Forgive me if I'm wrong, but in the `src/app.rs` `routes` function, the app routes are added with the `/api` prefix, meaning all routes have the `/api` prefix.

jondot:
That prefix update seems ok. The reason sometimes we go with and without `/api` is that Loco starters evolved from "API only" to many different types of full stack, between those server-side rendered full stack apps, which don't need the `/api` prefix because there is no separation between api and non-api.


In the most recent version we're returning a `Result<Response>` rather than a typed response (e.g. `pub async fn list(State(ctx): State<AppContext>) -> Result<Json<Vec<Model>>> {`) because it offers much more flexibility when returning different return types per some controller logic (e.g. conditionally returning Html vs JSON based on content type).

We don't have versioning on the docs, which we'd like to fix at some point, which is why the docs on master may not fit the docs which are online.
The best way to know if the docs in master fit reality, is to use Loco from master (apologies for that)

chmelevskij:
So to clear things up:
1. Keep the prefix changes removing `/api` routes, does it make sense to update the starter accordingly?
2. `Result<Response>` should stay, where is it supposed to be imported from? Does this need update to the starter also?
3. The docs are build from the tip of the master, but published loco version is behind?

At the moment the docs unfortunately are quite unusable for new comers 😭 to webdev/rust, the CRUD creation just fails silently and type mismatches don't help with experience 🫣 From user perspective feels like batteries included framework, but you need to duct tape few bits on and smash it with a hammer first 😆

P.S. Happy to help with the issues, just need a bit more guidance on the direction

jondot:
@chmelevskij yes, that's correct. apologies for the in-flux state, refactoring towards better DX and docs updates are a challenge (@kaplanelad is working on an automated solution for this)
In other news, v0.4.0 should resolve the confusion because it will align the released version with the docs, so things will be intuitive and working again.

We may need to consider to try working in a `dev` branch which will always symbolizes "next version" from now on to keep things clearer

chmelevskij:
will try speed running the tutorial over the weekend

chmelevskij:
@jondot, haven't tried full thing, but tried building small app.

This is the version I have which was generated with cli in my `cargo.toml`:
```
loco-rs = { version = "0.4.0", features = ["testing"] }
```

Quick check for `_health` and `_ping` and they still seem to be off as documented in https://loco.rs/docs/getting-started/guide/#building-a-crud-api

Screenshot 2024-04-21 at 12 03 53

chmelevskij:
The ` Result<Response> ` issue seems to be fixed though when using `scaffold` at least.

Also there is still issue when following `https://loco.rs/docs/getting-started/guide/#adding-a-second-model` there, is no `api` added.

Should I just update the paths to drop `api` in the docs then and remove those adjustments to types?

kaplanelad (approved):

ls-cam-smith:
I think this is good, I came to the same conclusion as a new user.

jondot:
Thanks!
I've just aligned the last pieces of the docs, please do reopen a new PR if you find more consistency issues!

PR #547, opened 2024-04-18
Adding the debug macro to the templates to have better errors.

Muxoid:
I'm new to Axum and Loco frameworks and encountered an error when trying to use a handler to return HTML instead of JSON. The issue stemmed from not adding Json as the final argument in the handler function, which the improved error message helped identify as you can see below.

I thought this would help with new users and DX, there might be a better way, let me know what you think. Maybe simply adding it to the docs might be sufficient.

[Doc](https://docs.rs/axum/latest/axum/attr.debug_handler.html)

image

### DIFF
--- a/starters/lightweight-service/src/controllers/home.rs
+++ b/starters/lightweight-service/src/controllers/home.rs
@@ -1,7 +1,9 @@
+use axum::debug_handler;
 use loco_rs::prelude::*;
 
 use crate::views::home::HomeResponse;
 
+#[debug_handler]
 async fn current() -> Result<Response> {
     format::json(HomeResponse::new("loco"))
 }
--- a/starters/rest-api/src/controllers/auth.rs
+++ b/starters/rest-api/src/controllers/auth.rs
@@ -1,3 +1,4 @@
+use axum::debug_handler;
 use loco_rs::prelude::*;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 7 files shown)


jondot:
This is great. Looks like there's no performance hit when compiling with `--release` which is fantastic.
I'm wondering how we can include this by default in all handlers, without letting the user manually type it above every single handler, is there such a way?

Muxoid:
I do not think it would be possible without changing the Axum code to make this default behavior. As it is the impl IntoResponse code that causes the complex error. The macro was created to solve the issue.

[Axum into_response.rs](https://github.com/tokio-rs/axum/blob/main/axum-core/src/response/into_response.rs)

jondot:
Looks like this should be useful by default, thanks!
Can you also add in the code generator templates? (remember you need to also include the import)

https://github.com/loco-rs/loco/blob/master/src/gen/templates/controller.t
https://github.com/loco-rs/loco/blob/master/src/gen/templates/scaffold/api/controller.t
https://github.com/loco-rs/loco/blob/master/src/gen/templates/scaffold/html/controller.t
https://github.com/loco-rs/loco/blob/master/src/gen/templates/scaffold/htmx/controller.t

Muxoid:
I have added it to the code generator templates as well.

kaplanelad:
Thanks @Muxoid Muxoid

kaplanelad (approved):

PR #555, opened 2024-04-24
Make health check more informative.

BWStearns:
I ran into an issue where it was difficult to tell which store was failing (redis or pg) so I figured this might be a helpful addition for others to make troubleshooting easier.

### DIFF
--- a/src/controller/health.rs
+++ b/src/controller/health.rs
@@ -6,31 +6,43 @@ use axum::{extract::State, response::Response, routing::get};
 use serde::Serialize;
 
 use super::{format, routes::Routes};
-use crate::{app::AppContext, redis, Result};
+use crate::{app::AppContext, db, redis, Result};
 
 /// Represents the health status of the application.
 #[derive(Serialize)]
 struct Health {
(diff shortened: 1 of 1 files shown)


kaplanelad:
Can you elaborate on why you need it? 
In the logs, you have an error message when the check fails, so you can understand the status of the checks from the log. 

You added tracing::info when it is a success; the check happens a lot, so this can manure your logs. why do you want to see in the logs if the check passes?

BWStearns:
> Can you elaborate on why you need it? In the logs, you have an error message when the check fails, so you can understand the status of the checks from the log.
> 
> You added tracing::info when it is a success; the check happens a lot, so this can manure your logs. why do you want to see in the logs if the check passes?

Good call on not logging it, but the motivation behind adding the details in the response was that I wasn't seeing the Redis check failure in my ECS logs. Maybe I missed it.

kaplanelad:
@BWStearns, how did you configure the healthcheck? from the load balancer? if yes you can validate only on the status code and not on the fields itself

PR #560, opened 2024-04-30
refactor auth middleware for supporting bearer, cookie and query

kaplanelad:
Extract JWT token from cookie

### DIFF
--- a/src/controller/middleware/auth.rs
+++ b/src/controller/middleware/auth.rs
@@ -36,7 +36,7 @@ use crate::{
 
 // ---------------------------------------
 //
-// JWT Auth extractor from bearer token
+// JWT Auth extractor
 //
 // ---------------------------------------
 
--- a/src/controller/middleware/auth.rs
+++ b/src/controller/middleware/auth.rs
@@ -19,16 +19,20 @@
 //!     format::json(TestResponse{ pid: auth.claims.pid})
 //! }
 //! ```
+use std::collections::HashMap;
 
 use async_trait::async_trait;
 use axum::{
-    extract::{FromRef, FromRequestParts},
+    extract::{FromRef, FromRequestParts, Query},
     http::{request::Parts, HeaderMap},
 };
 use axum_extra::extract::cookie;
 use serde::{Deserialize, Serialize};
 
-use crate::{app::AppContext, auth, errors::Error, model::Authenticable};
+use crate::{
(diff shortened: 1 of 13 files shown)


src/controller/middleware/auth.rs
```diff
@@ -165,3 +172,92 @@ where
...
+}
+
+/// Extract a token value from cookie
+///
+/// # Errors
+/// when token value from cookie is not found
+pub fn extract_token_from_cookie(parts: &Parts) -> eyre::Result<String> {
+    let jar = cookie::CookieJar::from_headers(&parts.headers);
+    Ok(jar
+        .get("token")
```
jondot, line 257: I would create some kind of unique Loco session ID key name.
in Rails it is: `_APPNAME_session`

and is also configurable:
```
config.session_store :cookie_store, key: '_somethingelse_session'
```

so we might need to reach into `AppContext` and then configuration for the cookie to grab the name of the cookie, and then when generating a new app give it a default name of `_APP_NAME_loco-session` (naming it `loco-session` to avoid conflict with rails)

src/controller/middleware/auth.rs
```diff
@@ -165,3 +172,92 @@ where
...
+        }
+    }
+}
+
+/// Extract a token value from cookie
+///
+/// # Errors
+/// when token value from cookie is not found
+pub fn extract_token_from_cookie(parts: &Parts) -> eyre::Result<String> {
+    let jar = cookie::CookieJar::from_headers(&parts.headers);
```
jondot, line 255: if possible, to take only `httponly` cookie or else put out an error. this will forcefully educate users to do the right thing

src/controller/middleware/auth.rs
```diff
@@ -165,3 +172,92 @@ where
...
+            }
+        }
+    }
+}
+
+/// Extract a token value from cookie
+///
+/// # Errors
+/// when token value from cookie is not found
+pub fn extract_token_from_cookie(parts: &Parts) -> eyre::Result<String> {
```
jondot, line 254: are we sure we want `eyre::Result` here?

kaplanelad: This function returns an error if the token cannot strip the value to get only the token. 
I prefer not to return an Option<String> and return an error with a message that explains where the error came from.

you prefer Result<String, String> instead of eyre::Result?

kaplanelad: changed

PR #570, opened 2024-05-03
Cache app context mechanism

kaplanelad:
This PR introduces a cache mechanism for the application context. By default, the memory cache layer is enabled, although users have the option to disable this feature.

## Changes Made
- App Hooks: Added a new cache hook within the App Hooks. This hook initializes a new instance of the in-memory cache engine with default configuration. Consideration should be given to moving this configuration into a centralized location.
- Cache Feature: The cache feature introduces a cache field under AppContext, allowing for the creation of specific cache implementations. The cache_inmem feature, enabled by default, includes both the cache feature and the [moka](https://crates.io/crates/moka) library.

## Discovery cach in memory libs:
- https://crates.io/crates/moka
- https://crates.io/crates/cached

Steps
 - [ ] Created a trait for a cache implementation to support different cache engines.
 - [ ] Memory caching is enabled by default (cache_inmem).
 - [ ] Consider adding more engines I.e Redis
 - [ ]  Added the cache feature, providing a cache field within AppContext and enabling the creation of specific cache implementations.
 - [ ]  Added the cache hook.
 - [ ] Tests.
 - [ ]  Consider refactoring the cache hook configuration into a centralized location for better organization.
 - [ ]  Enhance documentation for improved clarity and usability.
 - [ ] Think of strategy like ttl when insert value


jondot:
looks good
i think if the storage isn't optional, cache shouldn't be optional (we can toggle the cache drivers based on feature, but leave all the abstraction to be always inside)

src/app.rs
```diff
@@ -46,8 +48,11 @@ pub struct AppContext {
     pub config: Config,
     /// An optional email sender component that can be used to send email.
     pub mailer: Option<EmailSender>,
-    // Ab optional storage instance for the application
+    // An optional storage instance for the application
     pub storage: Option<Arc<Storage>>,
+    #[cfg(feature = "cache")]
+    // Cache instance for the application
+    pub cache: Arc<cache::Cache>,
```
jondot, line 55: consider to be exactly like storage, e.g. optional `Option<Arc<Storage>>`

kaplanelad:
@jondot, 
I have removed the feature flag and created a null driver for cache and storage like we spoke. 
Redis implementation will be in a different PR

jondot (approved):

PR #572, opened 2024-05-06
add repository field

szabgab:
To allow [Crates.io](https://crates.io/) , [lib.rs](https://lib.rs/) and the [Rust Digger](https://rust-digger.code-maven.com/) to link to it. See [the manifest](https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field) for the explanation.


jondot:
note that there are 2 "heads" for this repo
1 - workspace, which includes the loco crate, and the extras
2 - CLI, which is a standalone project 

I believe it is also needed to update the main workspace

szabgab:
If I am not mistake the two crates handle the meta data differently. One inherits from the root Cargo.toml the other one has the data itself. Probably it would be a better idea to use the same method in all the crates.

PR #573, opened 2024-05-06
Ilike dsl

yinho999:
- [x] Added in ilike, not_ilike, ilike_starts_with, ilike_ends_with, ilike_contains
- [x] Add in polymorphism for Postgres, MySQL and SQLite


yinho999:
@kaplanelad Which one do you think is a better approach? Using dsl traits for each engine, or using switch/match within the dsl function.

yinho999:
@jondot @kaplanelad, just a heads up that this code will cause a breaking change in the existing DSL code. Could you please double-check before merging? Thanks!

kaplanelad:
Hey @yinho999, sorry for the late response.

When you import `postgres::Postgres`, it depends on the feature flag in sea_orm. so in case one of the Loco users wants to Mysql and change the Postgres to MySQL, I think the code will not compiled. 

Do you consider this case and some cfg for this case?

yinho999:
@kaplanelad Thanks for your suggestion. Just want to double-check, do you mean using something like cfg_if to enable different code by the sea-orm feature flag for database variants?

yinho999:
@kaplanelad @jondot

Hello there,

I am looking for some suggestions regarding the feature flag approach in our DSL. Below is my current implementation using feature flags. However, this method has some issues.

If a user enables two different feature flags simultaneously, only one of the conditions will be compiled. Additionally, we need to update our CI/CD pipeline because running `cargo test --all-features` will lead to the same issue and cause the other two tests to fail.

```rust
    #[must_use]
    pub fn ilike<T: ColumnTrait, V: Into<String>>(self, col: T, a: V) -> Self {
        cfg_if::cfg_if! {
        if #[cfg(feature = "sqlx-postgres")] {
                with(
                    self.condition.add(
                        Expr::col((col.entity_name(), col))
                            .ilike(a)
                            .into_condition(),
                    ),
                )
            } else if #[cfg(any(feature = "sqlx-mysql", feature="sqlx-sqlite"))] {
                with(
                    self.condition.add(
                        Expr::expr(Func::lower(Expr::col(col)))
                            .like(a.into().to_lowercase())
                            .into_condition(),
                    ),
                )
            }
        }
    }
```

To address this issue, I can add the compilation time checks into the `lib.rs` or `main.rs` and update the CICD. However, I am still concerned about this approach since the compile-time check is loosely coupled with the logic.

```rust
// Compile-time checks for mutually exclusive features
#[cfg(all(feature = "sqlx-postgres", any(feature = "sqlx-mysql", feature = "sqlx-sqlite")))]
compile_error!("The 'sqlx-postgres' feature cannot be enabled with 'sqlx-mysql' or 'sqlx-sqlite' features.");

#[cfg(all(feature = "sqlx-mysql", feature = "sqlx-sqlite"))]
compile_error!("The 'sqlx-mysql' feature cannot be enabled with the 'sqlx-sqlite' feature.");
```

How does this approach look to you?

PR #578, opened 2024-05-09
Adding a custom claims Option<serde_json::Value> to the UserClaims struct.

Muxoid:
I want to be able to add custom roles to my JWT or any other security enhancements in the future. I think this would future-proof the JWT for the framework.

Let me know what you think.


jondot:
yes this is a great idea actually, please push forward when you can 👍

Muxoid:
I have fixed all the merge issues.

PR #580, opened 2024-05-10
Add MongoDB initializer

isaacdonaldson:
I added basic support for using a MongoDB database in the route handlers. No support for using Models or any of the other nice parts of using loco, but if someone wants to use MongoDB, they can choose to use this 🤷‍♀️ .


jondot:
Fantastic stuff!

PR #587, opened 2024-05-25
Update tour and guide doc resource paths to match current behaviour.

ls-cam-smith:
This resolves some issues when following the guide:
* example code should specify response as `Result<Response>`
* generated controllers do not fall under `/api/*`
* manual controller example updated to align with generated controllers to be clearer.

See #533, this continues @chmelevskij's work :)


ls-cam-smith:
Hi @jondot, @kaplanelad, this is another small PR for the guide docs. However I didn't review the playground section. I've marked this commit as Co-Authored by @chmelevskij because his PR came first.

PR #599, opened 2024-06-04
Add CC and BCC field to the mailers.

BWStearns:
Add a bcc field to Emails for the mailers.

### DIFF
--- a/src/mailer/email_sender.rs
+++ b/src/mailer/email_sender.rs
@@ -109,6 +109,10 @@ impl EmailSender {
             builder = builder.bcc(bcc.parse()?);
         }
 
+        if let Some(cc) = &email.cc {
+            builder = builder.cc(cc.parse()?);
+        }
+
         if let Some(reply_to) = &email.reply_to {
             builder = builder.reply_to(reply_to.parse()?);
         }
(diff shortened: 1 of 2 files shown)


BWStearns:
I haven't tested it end to end yet. Will comment when I have. Other than the existing email test in `email_sender` is there another place it should be tested in code?

BWStearns:
Unfortunately sendgrid [removed bcc](https://www.twilio.com/docs/sendgrid/ui/account-and-settings/mail#bcc) so I don't have a convenient way to test this end to end at the moment.

jondot:
Looks good. Should we also add cc for completeness?

PS I like using mailtutan (written in rust) or mailhog for development. It’s really nice.

BWStearns:
It at least doesn't error but I'm getting this in mailtutan:

```
[
    {
        "id": 1,
        "sender": "Someone contact@nativi.sh",
        "recipients": [
            "brianstestemail@gmail.com"
        ],
        "subject": "Welcome brian\n",
        "created_at": "2024-06-04 23:30:32",
        "attachments": [],
        "formats": [
            "source",
            "html",
            "plain"
        ]
    }
]
```

A quick search of their codebase says there's no cc or bcc in their codebase. Are these headers just not supported in mailtutan?

BWStearns:
image

Ok so CC works. I'm not seeing the bcc header but it looks like mailhog doesn't support it either. I think this looks good?

jondot:
yea lets go for it, thanks!

PR #610, opened 2024-06-10
bump version 0.6.0

kaplanelad:
Bump version to 0.6.0.

# Breaking Changes Upgrade Steps

### Change Description
Changed task args from BTreeMap to struct.

 ## Fix:
Replace task signature 
From:
```rust
async fn run(&self, app_context: &AppContext, vars: &BTreeMap<String, String>)
``` 

To:
```rust
n run(&self, _app_context: &AppContext, _vars: &task::Vars) -> Result<()>
```

PR: https://github.com/loco-rs/loco/pull/609

### Change Description
In order to avoid the over-used over-popular port of 3000 and avoid collisions of ports, cookies, sessions and app data, we suggest moving to [5150](https://en.wikipedia.org/wiki/5150_(album)).

PR: https://github.com/loco-rs/loco/pull/611


BWStearns:
You'll also want to update calls to `vars.get` in tasks to `vars.cli.get` like this.

```
    async fn run(&self, app_context: &AppContext, vars: &task::Vars) -> Result<()> {
        let refresh = vars
            .cli
            .get("refresh")
            .is_some_and(|refresh| refresh == "true");
```

Issue #600, opened 2024-06-04
Move away from default port of `3000` to `5150`


jondot:
In order to avoid the over-used over-popular port of 3000 and avoid collisions of ports, cookies, sessions and app data, we suggest moving to [5150](https://en.wikipedia.org/wiki/5150_(album)).

Also make sure `cargo loco start -p custom-port` works smooth and flawless if anyone wants to keep `3000`

dinosath:
can you assign it to me?

jondot:
Sure done, 
A few pointers:
* In addition for looking for the port in code, take a look at the CI jobs
* If you find any `3000` number in the wild which is not in a constant, try to refactor it into a constant
* the documentation probably will have a bit of changes, so mind those as well. we use a new tool @kaplanelad  developed for code snippets and code chunks called `snipdoc` (see its [configuration](https://github.com/loco-rs/loco/blob/master/snipdoc.yml)), see [how to install and use here](https://github.com/loco-rs/loco/blob/master/.github/workflows/docs.yml#L30)

PR #611, opened 2024-06-11
feat: change default port to 5150

dinosath:
Changed default port to 5150 according to #600


jondot:
Fantastic attention to detail, thanks!

PR #625, opened 2024-06-20
Use Rust-based tooling for SaaS starter frontend

omninonsense:
Migrate the SaaS frontend to a modern Rust-based toolchain, and enables TypeScript.

Overall these tools are more approachable in terms of configuration which makes the UX nicer, but their improved performance also improves DX.

Also, given this is a Rust framework, I thought it would be fitting to use and promote Rust-based web tooling!

## Adds TypeScript support

I think as this is a Rust repo, I probably don't have to give a sales pitch on the merits of having a type system! 😁 

## Replaces Vite with [Rsbuild](https://rsbuild.dev/)

Both are great tools, but Rsbuild is faster since most of it is built on top of Rust. Also, compared to other stuff in the ecosystem, it's a bit more "batteries included" when it comes to all the boring/infrastructural stuff.

Rsbuild is a wrapper around [Rspack](https://rspack.dev/) and [SWC](https://swc.rs/). Rspack is a drop-in webpack replacement, while swc is a babel alternative. Most webpack plugins will work with Rspack, but as a rule-of-thumb you generally don't need most of them these days (ones that operate on assets being an exception, but rsbuild/rspack provide those natively). SWC also has Rust plugins for a lot of the popular Babel plugins

Another benefit of Rsbuild is that it natively supports TypeScript out of the box. Its config can be written in TypeScript too, making it easier to understand and maintain the configuration.

If you're still not convinced, this will surely convince you: The mascot is gosh darn cute; matter settled.

## Adds [Biome](https://biomejs.dev/)

Biome is a linter and a formatter... Kinda like a 2-for-1 sale; who doesn't love a good bargain?

The [formatter](https://biomejs.dev/formatter) aims to be similar to Prettier, and they have a similar philosophy when it comes to configurability (ie both are opinionated). There are some differences; overall Biome tries to be more consistent.

When it comes to the [linter](https://biomejs.dev/linter), it has sensible defaults out of the box (the recommended rules). These lints are mostly around programming errors, safety, performance or accessibility issues. The only style lints it has enabled in the recommended set are ones that tend to make refactoring a bit more fiddly or error prone (but some extra style lints can be enabled if people want to).

Alas, they don't have a cute mascot.

## Other changes

Deletes the lock file, since users will probably wanna pull latest versions of the dependencies. I've also loosened the version ranges a bit in the package file to only pin major version.

Also deletes some of the files that are normally git-ignored in frontend/PWA projects, like the `dist` folder, and adds a few of the common CRA/Next-style `.env` files

---

> [!NOTE]  
> I've seen the 
> > Ideally, avoid making changes to the starters. Starters should have minimal implementations.
> 
> But these changes are relatively small incremental improvements. So, I took the liberty of opening a PR anyway. It wasn't a lot of work, so if you disagree with these, feel free to close the PR
>
> Alternatively, if people believe this is too much of a deviation, there's also [Farm](https://www.farmfe.org/). However, Rsbuild is a more holistic solution.


jondot:
This is a great idea actually. If there is a Rust-equivalent of a node toolchain or library, we want to use the Rust version.
I'll be happy to merge this.
Could you please add a way (through docs, or local project readme, or maybe  a `pnpm install-tools` command or something) so people can learn how to install all the extra tools? (biome, rsbuild)

omninonsense:
> This is a great idea actually. If there is a Rust-equivalent of a node toolchain or library, we want to use the Rust version. I'll be happy to merge this.

Yay! 😁 Glad to hear this, @jondot!

Regarding the failing tests: They are failing because I deleted the `frontend/dist/index.html` file. They were passing on my machine because I had the dist folder left over while configuring `rsbuild`. 

There's a few ways I can think of fixing the tests. Would like to get your input/preference. In order of simplicity:

1. Toggle the `must_exist` field to `false`:
https://github.com/loco-rs/loco/blob/aff0e74ac4c4d63728c1bcb1aca2a70fd802e933/starters/saas/config/test.yaml#L58
2. Reintroduce tracking of the `dist/` directory (slightly on the fence about this one, but I also think I understand why it was there to begin with)
3. Install node (probably easiest via [`fnm`](https://github.com/Schniz/fnm)) and pnpm and build the assets via `pnpm run build` before the tests run.

A variation of the third option is useful to have anyway, as it makes sure the frontend build system works. But to keep the PR smaller and on-topic, I'd rather we follow up on that.

So I'm happy for you to pick 1 or 2 and then we can follow up with another PR to discuss something for the third option. I'm getting overly excited and a bit ahead of myself, but who knows: maybe over time the Static Asset Middleware could even evolve to provide some functionality of [Rails's Asset pipeline](https://guides.rubyonrails.org/asset_pipeline.html) by wrapping `rsbuild`!

jondot:
I actually am happy you removed the dist folder, I think that is the correct thing to do.
Let's go with (1) since it's just test environment

omninonsense:
Toggled `must_exist: false` last night but forgot to update here. And I see you updated the branch (🙇‍♂️)

So, from my end should be ready to merge!

PR #627, opened 2024-06-22
Default binding to localhost to avoid firewall dialogues during development on macOS

trevoranderson:
On macOS, if you bind to an IP address (such as `127.0.0.1`) rather than `localhost`, then whenever you run the server after making changes you will receive a prompt from the system asking whether or not to allow connections. 
```
Do you want the application “locoapp-cli” to accept incoming network connections?
Clicking Deny may limit the application’s behavior. This setting can be changed in the Firewall pane of Network Settings.
```
This is a bit annoying, so this code changes the default to `localhost`. Previously, the default was `[::]` which is the same thing but triggers the firewall prompt.


jondot:
Hi! thanks for catching this. I feel it's a good change. It seems to me that this specific code change might nullify the configuration value (since it does not take into account the previous line which asks for the value from config).

Could you take a look at this case?

trevoranderson:
Moved the change to a place that looks more like the root cause. @jondot

jondot:
indeed, this is what Rails does
https://github.com/rails/rails/blob/de4d8744744acab2dd9db0683ccf784ea45810b2/railties/lib/rails/commands/server/server_command.rb#L221

jondot:
@trevoranderson could you check out the CI failures?

PR #636, opened 2024-07-09
feat: support logger file appender

tsailin-bit:
I would like to have the capability to save logs to a file and automatically roll them over on my server. When I came across the tracing-appender crate that provides this functionality, I attempted to incorporate it into the loco project. I'm not sure if it's appropriate, so I wanted to submit it as a pull request.

Here are the changes I made:

1. Added the tracing-appender crate as a dependency to the project.
2. Configured the rolling file logger with parameters such as the log file path, rolling strategy, and max file size.
3. Support non-blocking write mode

Please review my changes and let me know if it is appropriate to submit them as a pull request to the loco project. If you have any suggestions or modifications, please feel free to provide them. Thank you!


tsailin-bit:
> Exactly what I need. I having been putting this off (read: procrastinating), hoping that some kind soul would do the work for me...
> 
> But would this log to file _as well as_ displaying on stdout?

Yes, it will still output to stdout, but you can configure it to output only to a file or only to stdout. You can also configure it to output to both simultaneously.

kaplanelad:
@tsailin-bit can you please fix the CI?

tsailin-bit:
> @tsailin-bit can you please fix the CI?

yes, i will try fix!

tsailin-bit:
@kaplanelad Thanks for your reply，i fixed CI. EnvFilter did not show the expected behavior when applied to each layer separately, so it was adjusted to use a uniform EnvFilter pattern across all layers.

kaplanelad (approved):

kaplanelad:
thanks @tsailin-bit tsailin-bit

Issue #643, opened 2024-07-15
Add ability to provide fallback handler to AppRoutes


swlody:
## Feature Request

Currently, `controller::AppRoutes::to_router()` constructs an `axum::Router` with no fallback function. This means that it is not possible to have e.g. a custom 404 handler. It is possible to get around this by implementing `Hooks::after_routes()` and calling `Router::fallback()` on the provided `axum::Router`. However, the `after_routes()` callback is initiated after `AppRoutes::to_router()` is called https://github.com/loco-rs/loco/blob/7b716a73bf602c9a0aec95b0c72ac42791cf6305/src/boot.rs#L268-L269. This means that any middleware that was added is not provided to the fallback handler.

**Describe the solution you'd like**

Add a `add_fallback(fallback: Handler)` method to `controller::AppRoutes` so that a fallback handler can be added to the `axum::Router` and have all the relevant middleware applied to it.

This is potentially difficult to achieve since `axum::Router::fallback()` accepts an `axum::handler::Handler<T, S>` so `AppRoutes` would also need to be templated on `<T, S>` to be able to store the fallback on it.

**Describe alternatives you've considered**

Alternate solution is to move the call to `after_routes()` into `controller::AppRoutes::to_router()` so that it is actually applied immediately after the routes are registered. https://github.com/loco-rs/loco/blob/7b716a73bf602c9a0aec95b0c72ac42791cf6305/src/controller/app_routes.rs#L184-L190

This is however a potentially breaking change. Another problem with this method is that it would separate the location where `Hooks::after_routes()` is called from the location where `Initializer::after_routes()` is called. Also, `AppRoutes::to_router()` is not an async function while `Hooks::after_routes()` is (why?).

kaplanelad:
@swlod, 
you can use `after_routes` hook, which gets the axum router. 
this is the place you can add the fallback custom function

swlody:
> @swlod, you can use `after_routes` hook, which gets the axum router. this is the place you can add the fallback custom function

Maybe I'm missing something here. The problem that I am having using after_routes is that any fallback added in this function doesn't have any of the default middleware applied to it because it is called after the middleware is added to the router.

swlody:
Two related potential solutions:

Provide `Hooks::before_routes()` for the user to implement. Called after the initializers and before the router is created. The signature would look like this:

```rust
async fn before_routes(
    router: axum::Router<AppContext>,
    _ctx: &AppContext,
[...]

PR #646, opened 2024-07-20
Add Hooks::before_routes to give user control over initial axum::Router construction

swlody:
This function can be implemented by the user to fully control the construction of the axum router before it is handed back to loco. This is necessary in order to create a fallback handler that is covered by the loco middleware, as any fallback that is added in `after_routes` gets initialized *after* all the middleware in `to_router`. The default implementation of `Hooks::before_routes` results in the same behavior as before, but the user can override the implementation like so:

```rust
impl Hooks for App {
    // ...
    async fn before_routes(ctx: &AppContext) -> Result<AxumRouter<AppContext>> {
        // Anything that we want to add to the router before the middleware is initialized
        let router = AxumRouter::new().fallback(/* some fallback_handler */);
        Ok(router)
    }
    // ...
}
```

Resolves https://github.com/loco-rs/loco/issues/643


kaplanelad (approved):

jondot:
@swlody this is really great!
I always try to think in terms of process with the hooks, it keeps them more agile. So think:

"before lunch" (getting food, getting place to eat)
"having lunch" (eating)
"after lunch" (cleaning up, doing some exercise)

It doesn't always work to "force" things into that kind of shape and form but in this case, I think it might be.

I feel this PR (covering how to customize fallbacks, etc), is actually:

`before_routes(ctx: &AppContext) -> Result<AxumRouter>`

Where the user will

1. Customize / build / create their own initial version of Axum router, and finish at that
2. We (loco framework) take that router, and add the routes on it, while continuing to add the rest of the layers internally 

_if the user wants to add stuff "after", they can use after_routes, as you've already identified_

This keeps the complexity of "wait, after I built a new router, how to I add the routes?, what are the steps I need to follow to get to a fully working router? do I need to also call `after_routes` myself?" - this is a cognitive overhead that most users will have but "pro" users will probably manage (by viewing the source code or experimenting). We want to make sure *most users* actually do the right thing always (["the pit of success"](https://blog.codinghorror.com/falling-into-the-pit-of-success/)).

By offering the user just the thing they need ("e.g. use `before_routes`, create your own router and give it to us, we'll take it from there), we're ensuring a higher chance of success, and also we're making sure the framework (Loco) is always able to change things behind the scenes (order of initialization, adding more stuff into a router, etc.)

So to summarize:
1. Change `router` to `before_routes`, keep same signature, expect the user to only return a custom-initialized router (with fallback or etc.)
2. On the framework side, use this as a router construction method, and continue from there

Hope this wasn't too long, makes sense?

swlody:
Thanks for the detailed feedback @jondot. I've renamed to `before_routes` as you suggested and added a reference to the function in the docs.

kaplanelad:
thanks @swlody

PR #672, opened 2024-07-25
chore: add `publish = false` in the starter templates

haoxins:
I personally will add  `publish = false` in case it is published with an incident.
It will be more suitable to set  `publish` defaults to `false` as the starter template's default.


haoxins:
> @haoxins context will help here

added

PR #673, opened 2024-07-25
add get_or_insert function to cache layer

kaplanelad:
Adding the `get_or_insert` function to the cache driver for getting the option to get or insert in an easy way. 

I have also added to test_cfg a dummy app context for testing


src/cache/mod.rs
```diff
@@ -122,3 +156,31 @@ impl Cache {
...
+    #[tokio::test]
+    async fn can_get_or_insert() {
+        let app_ctx = tests_cfg::app::get_app_context().await;
+        let get_key = "loco";
+
+        assert_eq!(app_ctx.cache.get(get_key).await.unwrap(), None);
+
+        let result = app_ctx
+            .cache
+            .get_or_insert(get_key, &app_ctx, |_ctx: &AppContext| {
```
jondot, line 174: I assume there is no way to capture `ctx` right? (this is why its passed to the function)
If possible to get something like this:

```
.get_or_insert("key", async || {
     Users.find(..&ctx).await
})
```

this would be the most realistic. notice that 

1) `F` must be async because to be realistic people would probably make async db calls there
2) `ctx` if possible, should be captured to avoid passing it as a parameter just to "ship it" to `F`

Capturing `ctx` would probably require some concept of `sync` and `send`, so might require wrapping `ctx` in `Arc` and/or `Mutex`

src/cache/mod.rs
```diff
@@ -171,9 +172,7 @@ mod tests {
 
         let result = app_ctx
             .cache
-            .get_or_insert(get_key, &app_ctx, |_ctx: &AppContext| {
-                Ok("loco-cache-value".to_string())
-            })
+            .get_or_insert(get_key, async { Ok("loco-cache-value".to_string()) })
```
jondot, line 175: can you try a test where the async block captures something from app_ctx? like try a real DB call for example

kaplanelad: Added endpoint usage example + adding tests

examples/demo/src/controllers/cache.rs
```diff
@@ -11,14 +11,33 @@ async fn get_cache(State(ctx): State<AppContext>) -> Result<Response> {
...
+
 async fn insert(State(ctx): State<AppContext>) -> Result<Response> {
     ctx.cache.insert("value", "loco cache value").await.unwrap();
     format::empty()
 }
 
+async fn get_or_insert(State(ctx): State<AppContext>) -> Result<Response> {
+    let res = ctx
+        .cache
+        .get_or_insert("user", async {
```
jondot, line 23: something still seems off to me w/errors.

i believe, async block should just throw the regular Result from Loco (i.e. `CacheError` does not belong in that async block)

the block runs like any other "linear" code flow, only that in our case, the result of the block is cached. So the block can return a result or fail, like any other code flow. The failure should be for any *original* reason the block itself has failed.

a good way to look at it:

if you remove the cache completely, the code block should mostly stay the same

kaplanelad: got you, changed for supporting it

PR #682, opened 2024-08-01
 response from template

kaplanelad:
Adding the option to return template response in a simple way without initializer register or local files.
I have also changed the controller `dashboard` to `engine-view` 

- [x] Adding view engine tests
- [x] Testing controller


src/controller/views/mod.rs
```diff
@@ -26,6 +26,31 @@ impl<E> ViewEngine<E> {
...
+///
+/// # Errors
+///
+/// This function will return an error if building fails
+pub fn template<S>(template: &str, data: S) -> Result<String>
+where
+    S: Serialize,
+{
+    let mut tera = tera::Tera::default();
+    tera.add_raw_template("default", template)?;
```
jondot, line 50: I think we want: https://docs.rs/tera/latest/tera/struct.Tera.html#method.render_str

PR #706, opened 2024-08-16
FIX: hard coded /movies in SAAS templates view_edit

lerina:
minor glitch made new user experience unwelcoming.


jondot:
great catch!! thanks ❤️

PR #725, opened 2024-09-01
ci win32

jondot:


jondot:
current theory is that when we run "db migrate" from within a binary that was just compiled, it tries to compile a new binary (to include the new models and migrations) and run migrations, and that new binary tries to overwrite the current binary (from which the new process has been spawned) and this is not allowed in windows.

PR #736, opened 2024-09-05
Add `loco doctor --config`

### LINKED ISSUE #685: `loco doctor --config`
`loco doctor --config` should print out the currently *real* resolved configuration file: content + if possible path.
it should not run the actual doctor checks.

this is to accommodate a workflow where:

```
$ cargo loco doctor

< user bumps into an issue>
< user wants to troubleshoot, but everything looks fine >

$ cargo loco doctor --config

< user reviews their configuration, figures out they had a ".local" config which was resolved with different values, or 
   they had some kind of environment variable that influenced the final config values, etc. >
```

tetzng:
fixes #685


src/doctor.rs
```diff
@@ -166,3 +166,17 @@ pub fn check_seaorm_cli() -> Check {
         },
     }
 }
+
+/// Prints the configuration and its path.
+pub fn print_config(config: &Config) -> Result<(), Error> {
```
kaplanelad, line 171: This function is missing tests.
 It would be better to modify it so that it doesn't return a string directly, allowing us to use the `insta` crate to test the output. 
Alternatively, we could implement `fmt::Display` for the struct to handle printing the config details.

src/config.rs
```diff
@@ -73,6 +73,9 @@ pub struct Config {
     /// accessing `ctx.config.settings`.
     #[serde(default)]
     pub settings: Option<serde_json::Value>,
+
+    #[serde(skip)]
+    pub path: Option<PathBuf>,
```
kaplanelad, line 78: I think we can remove it, and print only the environment name

src/doctor.rs
tetzng: Thank you for the feedback! I have implemented `fmt::Display` for the `Config` struct.

src/config.rs
tetzng: Thank you! I have removed the path field and updated the code to print only the environment name.

src/doctor.rs
kaplanelad: can you add test?

tetzng: Thank you for your patience. I have added the test.

src/cli.rs
```diff
@@ -347,16 +351,21 @@ pub async fn main<H: Hooks, M: MigratorTrait>() -> crate::Result<()> {
         Commands::Generate { component } => {
             gen::generate::<H>(component.into(), &config)?;
         }
-        Commands::Doctor {} => {
-            let mut should_exit = false;
-            for (_, check) in doctor::run_all(&config).await {
-                if !should_exit && !check.valid() {
-                    should_exit = true;
+        Commands::Doctor { config: config_arg } => {
```
kaplanelad, line 354: config_arg => config

examples/demo/tests/cli_tests.rs
```diff
@@ -1,5 +1,25 @@
 #[test]
 fn cli_tests() {
     let t = trycmd::TestCases::new();
-    t.case("tests/cmd/*.trycmd");
+    t.case("tests/cmd/*.trycmd")
```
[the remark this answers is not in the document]
tetzng: I see your point, and I’ve removed the test.

src/cli.rs
tetzng: Thanks for the feedback. I used `config_arg` to avoid a naming conflict with the outer config, which is a `config::Config` struct. `config_arg` is a `bool`, so renaming it back to `config` would cause issues when referencing the `config::Config` struct. Keeping the names distinct is necessary to ensure the code functions correctly.

kaplanelad (approved):

PR #741, opened 2024-09-11
Rename demo:  `blo` -> `demo_app`

sectore:
## Why?

There are different code snippets in docs using `blo-cli` (for example [here](https://loco.rs/docs/getting-started/cli/) and [here](https://loco.rs/docs/getting-started/scaffold/)). 

For new users it's hard to understand `blo` is the name of the `examples/demo` app and is used to prefix `{my_app}-cli` commands.

## Changes:

- [x] Rename demo: `blo` -> `demo_app`
- [x] Run `snipdoc` to update code in docs
- [x] Cherry pick #747 to fix CI


jondot:
you're correct, this is a good catch

sectore:
@kaplanelad @jondot Just messed up the git history a little bit while rebasing, but it should not conflict anything by squash-merging this PR into `master`.

PR #777, opened 2024-09-22
Make `AppContext` extensible via traits

### LINKED ISSUE #522: Extensibility of AppContext
## Feature Request

The current [AppContext](https://github.com/loco-rs/loco/blob/master/src/app.rs#L37) does not allow users to extend the struct which forces the user to create a lot of workaround when working with extensions/initializers. 

We can create a thin wrapper which exposes the inner structure to the user and allow the user to add/modify the wrapper as they want.
```rust
use loco_rs::prelude::*;
struct LocalAppContext{
  app_context: AppContext
}
```
**Discord Discussion**
[Discord link](https://discord.com/channels/1184018317808259112/1217913658856701973)

mstallmo:
Closes #522 

This PR changes the `Hooks` trait to be generic over the `AppContextTrait` allowing end users to provide their own structs for application context. This allows for greater flexibility of end users to define the context of their web server as well as makes accessing user defined global application state smoother in various parts of the loco framework.

This change is implemented via two traits. The first being the object-safe `Context` trait and the second being the full `AppContextTrait`. The reason behind the split is that tasks and initializers are required to be object-safe and to pass in a generic type, that trait also needs to be object-safe. There are other places where the generic type is required to be `Clone` and/or `Default` making it impossible to implement in one trait.  So where object-safety is required the `Context` trait is used and everywhere else `AppContextTrait` is used.

Splitting the implementation into the `Context` trait and `AppContextTrait` also allows for the definition of a `create` method so the internals of the framework can construct the generic type with minimal disruption to the current code.


jondot:
@mstallmo this looks like a really interesting approach, thanks!
I activated CI, we have breaking changes.

I'm wondering how we can give users a pleasant path with all the breaking changes, as there are quite a few (considering AppContext is such a central subject).

mstallmo:
@jondot thanks for looking at my PR! I'll go ahead and make the changes to pass CI.

> I'm wondering how we can give users a pleasant path with all the breaking changes, as there are quite a few (considering AppContext is such a central subject).

Providing an easy path for users to upgrade has also crossed my mind. As a first step I would be happy to write up a section in the guide and a dedicated blog post detailing how to upgrade an existing app to this new way of implementing `AppContext`. From my experience with making the updates to the demo app the locations where users would have to make changes _should_ all be pretty much the same and mostly consist of adding type annotations where needed.

Because the existing `AppContext` struct implements the new context traits anywhere that `AppContext` is used explicitly by users should upgrade without any changes on their part.

---

Another option would be looking into writing a CLI tool or task to make the updates on the user's behalf. I don't have any prior experience in writing a tool like this but would be happy to work on one and take any suggestions the community might have on this topic.

kaplanelad:
Moving to draft until this pr is ready for review

PR #795, opened 2024-10-01
Add a note about what the options are for `loco new` in the Getting Started Doc

isaacdonaldson:
The Getting Started guide introduces the `loco new` CLI command and chooses the commands defaults. As a potential users first introduction to loco and `loco new`, I think it's useful to describe the other options to the choices and what they do/the effects they have. Added to a `infobox` so people can skip it if they are not interested.


kaplanelad:
The help and the field description are now part of the CLI.

Issue #801, opened 2024-10-06
"The Loco Guide" dose not work


Chris1717:
I'm new to loco and wanted to do the tour.

I installed the CLIs
```
cargo install loco-cli
cargo install sea-orm-cli
```

Started the project setup

```
christian@MBP14vohristian rust % loco new
✔ ❯ App name? · myapp
✔ ❯ What would you like to build? · SaaS app (with DB and user auth)
✔ ❯ Select a DB Provider · Sqlite
✔ ❯ Select your background worker type · Async (in-process tokyo async tasks)
✔ ❯ Select an asset serving configuration · Client (configures assets for frontend serving)

🚂 Loco app generated successfully in:
/Users/christian/Developer/rust/myapp

- assets: You've selected `clientside` for your asset serving configuration (remember to build your frontend in `frontend/`)
```
Changed into the "myapp" directory, and started the server:

`cargo loco start
`
Complies fine, but

```
2024-10-06T07:40:22.221004Z  INFO app: loco_rs::controller::app_routes: [Middleware] +log trace id environment=development
Error: Message("one of the static path are not found, Folder `frontend/dist` fallback: `frontend/dist/index.html`")
```

Loco does not start.

I would appreciate help.

lerina:
did you build your front end?
If you don't have a frontend (vue, react , ...) then you probably want to use 
tera serverside

Chris1717:
The guide does not ask to built a front end.
Do I get something wrong?

vitoraguila:
Yeah, also running in the same issue

ryane:
Yeah, building the frontend (following the instructions in the generated frontend/README.md) gets past the issue. Since the guide suggests that you should choose frontend serving, it should probably include (or at least link to) the instructions on building the frontend.

lerina:
The cli actually reminds you to do so:

- assets: You've selected `clientside` for your asset serving configuration (remember to build your frontend in `frontend/`)

chainyo:
The next command asks you to choose `--api`, `--htmx` or `--html` this could be updated too.

yonas:
Also, the API URL is now `http://localhost:5150/**api**/posts` not `http://localhost:5150/posts`

jondot:
closing since @isaacdonaldson have addressed it, thanks! 👍

PR #816, opened 2024-10-08
Adding some clarifying instructions to the getting started page

isaacdonaldson:
Looking at some feedback, it seems there is some confusion around some of the configuration options (https://github.com/loco-rs/loco/issues/801). This PR adds a couple sections to explain the differences in the getting started mini-tour depending on what option was chosen, as well as being explicit in needing to build the `frontend/` directory with the default.

Also worth noting: I changed the `127.0.0.1` host to be `localhost` to be consistent with the rest of the page (and work better on a copy/paste). I also added spaces under the `infobox` `div`'s because it was not highlighting the code in the backticks properly. This does cause the information icon to not be inline, but that is less important than being able to see the highlighting.


yonas (changes requested):
`/posts` should be `/api/posts`

isaacdonaldson:
> `/posts` should be `/api/posts`

Not the case all the time. I'm working on a PR that might address this, but currently, the routes that come with the starters are behind `/api` but the ones created by the generator do not have the `/api` prefix. At least this is true of version 0.9.

PR #818, opened 2024-10-09
Organize which routes are prefixed with "/api"

isaacdonaldson:
This PR aims to organize which routes are prefixed with `/api` and which are not. I moved where `/api` is prepended into the individual controllers and out of the routes section in the `app.rs` file, making all starters have the same basis. Changing the scaffold generation to prepend `/api` only when the `-k api` option is chosen, and not adding it for `-k htmx` and `-k html` (for the views) is easier to reason about. 

Before this change, scaffold generation would add a route as `/<resource_name>` to the `saas` starter, but as `/api/<resource_name>` to `rest-api` and `lightweight-service`. Instead, the new change would have `-k api` add new routes as `/api/<resource_name>` while `-k htmx` and `-k html` add them just as `/<resource_name>`. In this case, whether `/api` is added is based on whether you are using the `-k api` option, and not based on the starter template you chose.


jondot:
@isaacdonaldson i think this will unlock the bigger picture here: https://github.com/loco-rs/loco/issues/834

isaacdonaldson:
@jondot - Is there anything holding this up?

jondot:
@isaacdonaldson not at all, we've just made some changes to allow this to be part of a bigger picture of simplifying the starters

Issue #827, opened 2024-10-12
Custom Banner


leaderit:
It would be good to have an option for custom banner on start.
Extended information, diagnostics etc.

## The patch for the Feature is

[loco-rs+0.9.0.patch](https://github.com/user-attachments/files/17349376/loco-rs%2B0.9.0.patch)

WBRs,
Valerii

kaplanelad:
Hi! Thank you for the suggestion. 
To better understand your request, could you please provide an example of the custom banner output you're expecting? It would also be helpful to know:

- What specific fields or information would you like to include in the banner?
- Where you'd like each of those fields to be displayed.

This will help us better tailor the feature to your needs. Thanks!

kaplanelad:
adding `--no-banner` flag

PR #839, opened 2024-10-14
Add --no-banner flag

kaplanelad:
Add `--no-banner` flag to allow disabling the banner display.

related issue: #827


examples/demo/examples/start.rs
```diff
@@ -14,6 +14,6 @@ async fn main() -> loco_rs::Result<()> {
         port: boot_result.app_context.config.server.port,
         binding: boot_result.app_context.config.server.binding.to_string(),
     };
-    start::<App>(boot_result, serve_params).await?;
+    start::<App>(boot_result, serve_params, true).await?;
```
jondot, line 17: i would invert the bool meaning
the regular action is that we show banner
the special action is that someone requests a no-banner
so i would name the flag `hide_banner` with default false

kaplanelad: Changed.

jondot (approved):

Issue #828, opened 2024-10-12
Error: One of `kind`


nagashi:
**Description**

My OS is a Mac, which is up to date.  Rust version is 1.81.0 (not nightly), loco-cli is version 0.2.9 with loco-rs version 0.9.0.  My dependencies are as follows:
[dependencies]

loco-rs = { version = "0.9.0" }
migration = { path = "migration" }

serde = { version = "1", features = ["derive"] }
serde_json = "1"
tokio = { version = "1.33.0", default-features = false }
async-trait = "0.1.74"
tracing = "0.1.40"
chrono = "0.4"
validator = { version = "0.16" }
sea-orm = { version = "1.0.0", features = [
  "sqlx-sqlite",
  "sqlx-postgres",
  "runtime-tokio-rustls",
  "macros",
] }

axum = "0.7.5"
include_dir = "0.7"
uuid = { version = "1.6.0", features = ["v4"] }
tracing-subscriber = { version = "0.3.17", features = ["env-filter", "json"] }

# view engine i18n
fluent-templates = { version = "0.8.0", features = ["tera"] }
unic-langid = "0.9.4"
# /view engine

[[bin]]
name = "myapp-cli"
path = "src/bin/main.rs"
required-features = []

[[bin]]
name = "tool"
path = "src/bin/tool.rs"
required-features = []

[dev-dependencies]
serial_test = "3.1.1"
rstest = "0.21.0"
loco-rs = { version = "0.9.0", features = ["testing"] }
insta = { version = "1.34.0", features = ["redactions", "yaml", "filters"] }

When running the following command, cargo loco generates controller articles, I receive the following:
Error: Message("Error: One of `kind`, `htmx`, `html`, or `api` must be specified.") 

**To Reproduce**
I'm following The Loco Guide: https://loco.rs/docs/getting-started/guide/  
When trying to plug this into an articles controller is when I receive the error message.

**Expected Behavior**

I expected to see the following:
$ cargo loco generate controller articles
added: "src/controllers/articles.rs"
injected: "src/controllers/mod.rs"
injected: "src/app.rs"
added: "tests/requests/articles.rs"
injected: "tests/requests/mod.rs"

**Environment:**

**Additional Context**

Include additional context about the issue in this section. For instance, share insights into the bug's discovery process or any hypotInclude additional context about the issue in this section. For instance, share insights into the bug's discovery process or any hypotheses you may have regarding the root cause or specific aspects where Loco may be malfunctioning.

[...]

PR #840, opened 2024-10-14
docs: adding -k to missing parts

kaplanelad:
related issue: https://github.com/loco-rs/loco/issues/828


docs-site/content/docs/getting-started/guide.md
```diff
@@ -320,7 +320,7 @@ A model in Loco represents data *and* functionality. Typically the data is store
 Let's create a new model called `Article`:
 
 ```sh
-$ cargo loco generate model article title:string content:text
+$ cargo loco generate model article title:string content:text -k api
```
jondot, line 323: should we do `--api` instead?

kaplanelad: Done in all referances

Issue #796, opened 2024-10-03
Panic in `insert_many` with many nullable fields when seeding data


CandleCandle:
**Description**

I think this is more a problem with [seaorm](https://github.com/SeaQL/sea-orm) than Loco, however, posting here gives the work-around exposure in the place where it is relevant. It may be sensible to implement a work-around in Loco.

https://github.com/SeaQL/sea-orm/issues/1407

**To Reproduce**

I have a table with several nullable columns. Loco takes the single yaml file as input and converts it to an `insert_many` call as part of the `seed`.

Observe that the entry at ID 2 does not set a value for `nullable1`, nor does the first entry set a value for `nullable2`.

Sample failing yaml:
```yaml
- id: 1
  something: some value
  nullable1: value
- id: 2
  something: some value
  nullable2: value
```

**Expected Behavior**

The above database entries to be created with missing fields being `null`.

**Environment:**

**Additional Context**

**Work-Arounds**

There are two possible work-arounds:
1. Each yaml file to include empty values for unused fields:
```yaml
- id: 1
  something: some value
  nullable1: value
  nullable2:
- id: 2
  something: some value
  nullable1:
  nullable2: value
```

2. group your inserts into separate files, one file per distinct set of nullable fields with non-null values. This separates it into many calls to `insert_many`

File 1:
```yaml
- id: 1
  something: some value
  nullable1: value
```

File 2:
```yaml
- id: 2
  something: some value
  nullable2: value
```

jondot:
Thanks for the report, nice catch!
We'll think of a solution for this

jondot:
Looks like this only mean we need to work around and insert one by one

tyt2y3:
Thank you everyone. This has been resolved in SeaORM 1.1.3 https://github.com/SeaQL/sea-orm/issues/1407

PR #845, opened 2024-10-14
fix nullable fields when seeding data

kaplanelad:
When using insert_many, it's expected that the schema for each row is the same. For more details, refer to the [issue description here](https://github.com/loco-rs/loco/issues/796).

As a temporary workaround until SeaORM addresses the [related issue](https://github.com/SeaQL/sea-orm/issues/1407), I've modified the logic to insert the seed data one row at a time.


tyt2y3:
Just a heads up, I've released the fix in SeaORM 1.1.3. So this workaround can probably be reverted in the next minor release

Issue #773, opened 2024-09-20
Optimizing SQLite for servers


double-daruma:
**Description**

I think it would be cool that when using SQLite that the defaults that loco uses work well even for a production deployment. Since many smaller apps with only like 100 users or so can run in production with SQLite. This Blog Post has a good summary of how to use SQLite for server applications: https://kerkour.com/sqlite-for-servers

In particular setting pragmas:
```
PRAGMA journal_mode = WAL;
PRAGMA busy_timeout = 5000;
PRAGMA synchronous = NORMAL;
PRAGMA cache_size = 1000000000;
PRAGMA foreign_keys = true;
PRAGMA temp_store = memory;
```
And using `BEGIN IMMEDIATE` should be possible to implement. The part about using two different connections for read and write could be harder.

Thank you for your consideration.

Sillyvan:
I agree that we need to change the SQlite defaults in loco. The defaults of sqlite are old and cant be changed by them due to compatibility. Instead of relying on that website i would prefer to just copy the new rails adapter for sqlite because that one has been optimized recently.

https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb

Sillyvan:
Im reading a lot of code because i would really like to contribute to this project. But i just really cant figure out where to add them. Im getting the feeling that i would need to implement a bunch of things for the Sqlite options. similar to the additional postgres methods.

Maybe this can help? https://github.com/SeaQL/sea-orm/issues/2347

Sillyvan:
@double-daruma we merged some optimizations.

BEGIN IMMEDIATE is something i never saw anyone talk about before and it might be more of a thing for SeaORM.
The two connections are also a pain to implement and only every matter with huge loads. Is this enough for now?

PR #848, opened 2024-10-14
Provide good default PRAGMAS for SQLite

Sillyvan:
Addressing Issue https://github.com/loco-rs/loco/issues/773

### What has been done
Just a very basic way of setting PRAGMAS on connect.
Defaults inspired by the [official sqlite ruby adapter.](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb)

### Why
SQLite is very much a DB that can be used in production nowadays. Due to age and importance of compatibility those settings are not default. Especially `journal_mode = WAL` will improve the experience a lot by not be blocking on reads or writes anymore.

If there is anything missing from my PR or if there is any concerns or other opinions im happy to hear about it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 todo.txt
 examples/demo2
 *.sqlite
+*.sqlite-wal
+*.sqlite-shm
 
 # IDE config files
 .idea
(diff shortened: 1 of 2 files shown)


ryands17:
Is it possible to configure these values? I have a project where I can't use foreign keys so I'd like to just disable that.

kaplanelad:
> Is it possible to configure these values? I have a project where I can't use foreign keys so I'd like to just disable that.

No, this option is not available at the moment.

PR #853, opened 2024-10-15
improve app name validation

kaplanelad:
Following the finding in fuzzy generator template findings, I took the [cargo validation](https://github.com/rust-lang/cargo/blob/master/crates/cargo-util-schemas/src/restricted_names.rs#L47) and moved it to our validation


loco-cli/Cargo.toml
```diff
@@ -48,6 +47,7 @@ cfg-if = "1.0.0"
 dunce = { version = "1.0.4" }
 strum = { version = "0.26", features = ["derive"] }
 colored = "2"
+unicode-xid = "0.2.6"
```
jondot, line 50: i think we can drop checking for unicode by the fact that we don't allow unicode in app name.
its reasonable to just say app names are alphanumeric (ASCII)

PR #869, opened 2024-10-19
allow dash in app-name

joshka:
I like to have Rust projects that have dashes in their name and was annoyed that this was not possible.
This changes the lib name place holder to be snake case, and adds a new app name placeholder that is not modified.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["macros", "xtask", "loco-extras"]
+members = ["macros", "xtask", "loco-extras", "loco-cli"]
 exclude = ["starters"]
 
 [workspace.package]
--- a/loco-cli/Cargo.toml
+++ b/loco-cli/Cargo.toml
@@ -1,5 +1,3 @@
-[workspace]
-
 [package]
 name = "loco-cli"
 version = "0.2.9"
@@ -48,6 +46,7 @@ cfg-if = "1.0.0"
 dunce = { version = "1.0.4" }
 strum = { version = "0.26", features = ["derive"] }
 colored = "2"
+heck = "0.5.0"
 
 [dev-dependencies]
 trycmd = "0.14.19"
--- a/loco-cli/src/bin/main.rs
+++ b/loco-cli/src/bin/main.rs
@@ -1,6 +1,7 @@
 use std::{env, path::PathBuf};
 
 use clap::{Parser, Subcommand};
+use heck::ToSnakeCase;
 use loco_cli::{
     generate::{self, AssetsOption, BackgroundOption, DBOption},
     git, prompt, CmdExit,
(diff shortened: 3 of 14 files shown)


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["macros", "xtask", "loco-extras"]
+members = ["macros", "xtask", "loco-extras", "loco-cli"]
```
joshka, line 2: I wasn't entirely sure why loco-cli was not included in the main workspace, and added it so that rust-analyzer works when opening the repo.

joshka:
Test failure is unrelated to this change (windows newline diffs in the mailer template generation insta tests).

kaplanelad:
Hey @joshka 
Thank you for this PR! I 

My main concern is that merging this PR could introduce breaking changes for users who have not updated to the latest version of the CLI.

To check this, I cloned your repository and built it using the old version of the Loco CLI (the current released version). I ran the following command, which points to your repo (only for the starters changes), while using the current latest version of the CLI (without your changes):

```sh
STARTERS_LOCAL_PATH=/tmp/loco-rs/loco loco new
```

When using `myapp` as an example application name, the cargo check command failed.

```
error[E0433]: failed to resolve: use of undeclared crate or module `appname`
 --> src/bin/main.rs:2:5
  |
2 | use appname::app::App;
  |     ^^^^^^^ use of undeclared crate or module `appname`
  |
help: there is a crate or module with a similar name
  |
2 | use AppName::app::App;
  |     ~~~~~~~

error[E0433]: failed to resolve: use of undeclared crate or module `appname`
 --> src/bin/tool.rs:2:5
  |
2 | use appname::app::App;
  |     ^^^^^^^ use of undeclared crate or module `appname`
  |
help: there is a crate or module with a similar name
  |
2 | use AppName::app::App;
  |     ~~~~~~~
```

Let me check if we can avoid these breaking changes and prevent the need for users to update the Loco CLI.

joshka:
That's not something I considered, but makes sense. I'm not sure I can see an obvious way to get around that - perhaps add an extra patten for "AppName" that replaces it with "please update your loco-cli" would work?

kaplanelad:
Yes, I tend to agree.

I’m wondering if, with these breaking changes, we should consider the following for the next braking changes releases:

1. Move the `generator.yaml` to be part of the CLI binary rather than the template.
2. Implement version validation in `loco-cli` so users receive a clear error message: "This template requires loco-cli version X or above. Please run cargo install loco-cli to update."

joshka:
One other approach might be to make the starters included in the cli and avoid doing the git clone. A larger change obviously, but if you can adjust your release cadence (automatically do a cli point release on change) that could work too.

kaplanelad:
Sounds like a great suggestion! I'll work on moving our CLI into the crate and set the version in the starters folder. I'll let you know the results.

joshka:
The other benefit of doing that is that if there's ever a bug in a template version, you can just downgrade the cli and use that while waiting on a fix. That's inherently user friendly while also enabling devs to move fast and break things.

kaplanelad:
great, I'll prepare a branch (based on your changed) and share with you the result

kaplanelad:
Hey @joshka. 
This is solved in the new loco cli as an impeding template.
Thanks!

joshka:
Thanks!

Issue #881, opened 2024-10-20
reduce monorepo footprint


jondot:
In order to make build folder take less space, and to ease up on rust-analyzer, we'd like to

* [ ] Move `loco-cli` to its own independent repo (this was part of the plan until it stabilizes)
* [x] Move or remove `macros`, this was an experimental project that did not take off
* [x] Move `loco-extras` to its own repo, make sure that we still have interconnectedness between `loco` and `loco-extras` in terms of breakages
* [ ] refactor + simplify `xtask` bump version process https://github.com/loco-rs/loco/pull/893
* [ ] add a 'publish' step for subcrates, then main crate via `xtask`

joshka:
I really like that you have a mono-repo - it's a plus when you can just make changes and have them coherent with the rest of the repo. This is fairly small project all things considered - larger by at least two orders of magnitude git repos are still good for monorepos IMHO.

What's the problem with r-a that you're hitting?

jondot:
@joshka you're right, better to keep a monorepo. right now i'm hitting a couple of limits:
* `target/` folder size can reach 10-40GB on occasions
* rust-analyzer becomes sluggish as we're growing the list of members in the workspace

nevertheless, I've just added a subcrate: `loco-gen` because I had no other choice. For this instance, `loco-gen` will be the home of all things generators, including -- and more importantly -- generator testing, and generator testing was so heavy and filled with side-effects I had to isolate it in the form of a separate crate.

so looks like we're going to grow the amount of crates in the workspace any way, but then we'll remove some unneeded ones.

jondot:
splitting remaining work into:

* https://github.com/loco-rs/loco/issues/913
* https://github.com/loco-rs/loco/issues/914

PR #883, opened 2024-10-20
refactor: move loco-rs crate under loco-rs folder

joshka:
this cleans up the root and makes it easier to focus on just one crate at a time

--- 

This is a PR that's sort of a RFC - I won't be offended if you close it as not interested. 
It may be related some ideas in little to https://github.com/loco-rs/loco/issues/881


jondot:
Thanks! this is a nice take actually.
We're going to dedicate `0.12.0` under the theme of "removing, simplifying" (https://github.com/loco-rs/loco/issues/881)

The idea is that I've put everything bunched up, so that we can see and in that regards "feel" our deps, it makes rust-analyzer's work a bit heavier, it makes the target folder a big bigger, and it makes shipping a release a bit slower. 
So we're now putting things into their natural places to simplify and make things more lightweight and easier.

For example: the `macros` project, did not take off (we thought giving everything a macro for better DX, but that turned out to be un-Rustful)
The `extras` project evolve in a different speed, and is more "static", and so should be out of the workspace, but *closely tested* against Loco versions
And @kaplanelad is now experimenting with unifying `loco-cli` and `loco` into a single binary+lib crate, so that we can couple new project generation with its testing, and with current versions of the framework (just like Rails does).

We'll see how the simplification effort goes and definitely want to look at this PR again.

PR #909, opened 2024-10-26
Fix truncate not actually truncating

Ken-vdE:
This fixes an issue that causes postgres database to not actually truncate, so Id column auto-increment does not get reset to 1 and new models continue to use Id from existing increment.
This truncate query is actually faster and this fixes an issue where some tests fail (when using postgres db) because snapshot thinks new entities should have Id 1 when they do not.

Now, I'm very new to both Rust and Loco, so please inspect this code carefully and make sure I haven't done anything stupid.


jondot:
Thanks for this PR!
Your code looks good.
I think the semantics of truncate is to not restart identity (see rails: https://github.com/search?q=repo%3Arails%2Frails%20build_truncate_statement&type=code)
But the test errors I believe come from a change in how seeding is done (where seeding provides explicit ID), we're still trying to manage to get that ID to take effect in a different PR.
Either way moving to truncate is speedier, I'll leave that PR open until we locate the specific fix for seeding (in a different effort) and then see how to incorporate

Ken-vdE:
Sounds good.
Yes, truncate is speedier. Furthermore, calling truncate on a database does something very different than running a delete-all query. So, if you're going to name the function 'truncate', then it should actually run truncate. Otherwise the function should probably just be named 'delete_all'.

jondot:
you're right.
thinking ahead of incorporating this, I'm thinking about the topic of escaping and all kinds of edge case.
could you try using `sea-query` directly? looks like they have facilities to compose queries in a safe way: https://github.com/SeaQL/sea-query?tab=readme-ov-file#table-truncate

Ken-vdE:
@jondot I've changed the hardcoded query string to a sea_query built string.\
It's a little more verbose but safer. This escapes the table name.\
Is this what you were looking for? If so, please triple check the code and make sure it's correct.

kaplanelad:
The CI checks have failed. If this is still relevant, resolve the CI issues and reopen the PR.

PR #910, opened 2024-10-26
Fix tests being generated with wrong endpoints.

Ken-vdE:
Because both [loco-gen/src/templates/controller/api/controller.t](https://github.com/loco-rs/loco/blob/master/loco-gen/src/templates/controller/api/controller.t) and [loco-gen/src/templates/scaffold/api/controller.t](https://github.com/loco-rs/loco/blob/master/loco-gen/src/templates/scaffold/api/controller.t) generate the routes as plural, the generated tests should reflect that. 
([loco-gen/src/templates/controller.t](https://github.com/loco-rs/loco/blob/master/loco-gen/src/templates/controller.t) however, does not for some reason? Maybe because that not necessarily reflects an 'entity')

Also, I'm not sure why these duplicate files exist. 
Could both the controller and test templates (/templates/controller, /templates/scaffold & /templates) not be combined into single files?
Currently they do the something similar but they are generated with slightly different code.

I'm very new to Loco, so please verify I haven't done something wrong here.

### DIFF
--- a/loco-gen/src/templates/controller/api/test.t
+++ b/loco-gen/src/templates/controller/api/test.t
@@ -12,6 +12,19 @@ use {{pkg_name}}::app::App;
 use loco_rs::testing;
 use serial_test::serial;
 
+#[tokio::test]
+#[serial]
+async fn can_get_{{ name | plural | snake_case }}() {
+    testing::request::<App, _, _>(|request, _ctx| async move {
+        let res = request.get("/{{ name | plural | snake_case }}/").await;
+        assert_eq!(res.status_code(), 200);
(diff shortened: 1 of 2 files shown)


jondot:
Thanks, just to answer your question, initially the 'controller' generator created a "hello world" controller, without taking any fields or parameters. Today, this controller is used mostly to generate "pages" or single routes (as opposed to scaffold which generates a full stack CRUD).
You identified a place where I believe we missed the refactor, thanks!

We keep these generators separate (although they are almost identical) to allow for freedom of operation in each, for now.

PR #922, opened 2024-10-28
fix wrong injection in migration/lib.rs

kaplanelad:
Following the changes in https://github.com/loco-rs/loco/pull/887, the migration was not correctly injected into migration/lib.rs. 
See error example here: https://github.com/loco-rs/loco/actions/runs/11548816568/job/32140819625

To fix this fast, I updated the injection rule by moving it above the current injection point and added a newline comment (//) as a marker for the injection rule.

### DIFF
--- a/loco-gen/src/templates/model.t
+++ b/loco-gen/src/templates/model.t
@@ -7,7 +7,7 @@ skip_glob: "migration/src/*_{{plural_snake}}.rs"
 message: "Migration for `{{name}}` added! You can now apply it with `$ cargo loco db migrate`."
 injections:
 - into: "migration/src/lib.rs"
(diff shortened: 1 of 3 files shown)


jondot:
Thanks fixed via https://github.com/loco-rs/loco/commit/7bb5e4c79ec0ab323c64a45128a37d8df9ebceb6

ABruel:
> Thanks fixed via [7bb5e4c](https://github.com/loco-rs/loco/commit/7bb5e4c79ec0ab323c64a45128a37d8df9ebceb6)

Couldn't find the pr for this change, so I'll comment here. `inject-below` doesn't seem to work, should probably be `inject-above`. See #949.

PR #924, opened 2024-10-28
xp: optimize tool

jondot:


jondot:
resolved as part of new loco new

PR #927, opened 2024-10-28
Use a single loco-rs dep for a whole project

darricheng:
With this change, users would only need to update the `loco-rs` dep in a single place versus to up to three previously for the saas and rest-api starters.


jondot:
Fantastic. I will need to first test the release script 
Will merge once I’m done today or tomorrow

jondot:
What are your thoughts about disabling default features? 

(defaults are `default = ["auth_jwt", "cli", "with-db", "cache_inmem", "bg_redis", "bg_pg"]`)\

Is it possible that for example `saas` is not getting `with-db`? although I see tests are passing, so its a bit strange

Issue #920, opened 2024-10-27
Add hot reload for tera templates


Swoorup:
## Feature Request

Can we add hot reload mechanism for tera templates.

**Is your feature request related to a problem? Please describe.**

N/a

**Describe the solution you'd like**

Editing tera .html files should result in immediate changes without compiling source. See also https://www.reddit.com/r/rust/comments/1f2rbe7/livereload_using_tera_in_axum

**Describe alternatives you've considered**

N/A

jondot:
Thanks for the feedback!
We actually experimented with this (see [https://github.com/loco-rs/loco/blob/master/src/controller/views/engines.rs#L57](https://github.com/loco-rs/loco/blob/master/src/controller/views/engines.rs#L57))
Then we commented it out because it meant having `unsafe` code in Loco.

We want to bring it back, but only if we can make sure that it's only enabled in Debug builds, and no `unsafe` code in the Loco framework.

Right now we're recommending `cargo-watch` until we can find a good idea to accomplish the above.

jondot:
I've implemented #977 as the simplest way to achieve a reload effect.
let's take your #930 and turn it into a live-reload only integration, if you can shape it in a form of a middleware that would be the best option for us 👍 
closing this one

PR #930, opened 2024-10-29
[WIP] Hot reloading

Swoorup:
Awesome project btw. I kinda started on https://github.com/loco-rs/loco/issues/920 and wanted your input. 

- Semi working hot reloading without unsafe. 
- Need to also attach the tower-livereload to make it fully automated. 

It looks like the `tera` member in the `TeraView` struct needs to be made private, to not expose the leaky details about it being either in a Mutex in `debug` vs not wrapped in `release` builds.

Perhaps we could add 2 methods, 
* `fn with_tera<'s, A>(&'s mut self,  f: impl FnOnce(&'s tera::Tera) -> A ) -> A`
* `fn with_tera_mut<'s, A>(&'s mut self,  f: impl FnOnce(&'s mut tera::Tera) -> A ) -> A`

Bit also unsure about wiring up the tower-livereload without breaking existing interfaces. I did a roughed up patching within my own code and it works without issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,6 +120,9 @@ object_store = { version = "0.11.0", default-features = false }
 # cache
 moka = { version = "0.12.7", features = ["sync"], optional = true }
 
+notify-debouncer-full = { version = "0.4.0" }
+tower-livereload = { version = "0.9.4" }
+
 # Scheduler
 tokio-cron-scheduler = { version = "0.11.0", features = ["signal"] }
 english-to-cron = { version = "0.1.2" }
--- a/src/controller/views/engines.rs
+++ b/src/controller/views/engines.rs
@@ -1,5 +1,16 @@
-use std::path::Path;
+use std::{
+    path::Path,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
 
+use tower_livereload::LiveReloadLayer;
+
+use notify_debouncer_full::{
+    new_debouncer,
+    notify::{EventKind, RecommendedWatcher, RecursiveMode},
+    DebounceEventResult, Debouncer, FileIdMap,
+};
 use serde::Serialize;
 
 use crate::{controller::views::ViewRenderer, Error, Result};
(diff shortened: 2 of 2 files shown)


jondot:
Hi! happy you decided to contribute!

I'm wondering if we can split this into:

* reloading content from disk
* live-reload (triggering browser refresh) -- which will probably be solved by some kind of middleware glue (maybe..?)

There is a good chance a simplification can be made for reloading from disk - by just reading a file, and then rendering using a template string (instead of arc/mutex/ etc.). It can make a simpler (yet less clever) code, since we don't much care about perf in debug builds. WDYT?

jondot:
Moving this one to 0.14, if you can refactor it into a middleware that only does live-reload that would be perfect!

Swoorup:
Just back from holidays, I'll try to see if I can come up with something simpler this or next week. Think most of the reload logic could probably live in the user side, if we somehow expose methods like `full_reload` and paths in the `TeraView`

kaplanelad:
@Swoorup @jondot,

I have a few notes regarding this PR, live reload functionality, and some related feedback from my app.
 Here's the bottom line: I believe this feature should be gated behind a feature flag

First (not actually relevant for this PR, but relevant to hot reload)
In the Initializer view engine, there are two separate logic flows for registering functions into Terra.
Refer to this [line](https://github.com/loco-rs/loco/blob/master/loco-new/base_template/src/initializers/view_engine.rs#L22).

Currently, if I want to register additional functions or filters, I need to implement them in two different places. This is not intuitive and can lead to problem in release (and the test should to **catch this problem**). If this was done solely for hot reload support, I think we should consider a different approach for its implementation.

About this PR:
Hot reload functionality is relevant only during development. As part of our effort to minimize dependencies and reduce compilation time, the following crates should not be included in release builds:
```
notify-debouncer-full = { version = "0.4.0" }
tower-livereload = { version = "0.9.4" }
```

And regarding ` #[cfg(debug_assertions)]`
Using `#[cfg(debug_assertions)]` introduces issues in tests because it leads to behavior that differs between development and production builds. Here's an example test that fails due to this discrepancy:
```rust
pub fn test() -> String {
    #[cfg(debug_assertions)]
    let res = "#[cfg(debug_assertions)]";
    #[cfg(not(debug_assertions))]
    let res = "#[cfg(not(debug_assertions))]";

    res.to_string()
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_debug_assertions() {
        assert_eq!(test(), "#[cfg(not(debug_assertions))]");
    }
}
```

kaplanelad:
We now reload content from disk.

PR #952, opened 2024-11-02
fix: codegen generating migrations that apply backwards

### LINKED ISSUE #949: Can't scaffold references. Migrations applied in the wrong order.
**Description**

I simply can't find a decent way to generate a scaffold with references. 

`$ cargo loco generate scaffold post title:string! content:string! author:references --api`

This creates and updates all files and for some reason automatically runs the posts migration even though the generated files make no sense. There's no `authors` table, I want it to reference the users table. I *could* use `user` instead of `author` but that then brings the problem of creating multiple references to the same table like in trying to create `updated_by` and `created_by` columns.

**To Reproduce**
Tried several things, they either don't work at all or require too much work on my part for something that could easily be generated. All apply migrations in the wrong order or roll them back in the wrong order.

Everything was run on this project:
```
❯ loco new
✔ ❯ App name? · test
✔ ❯ What would you like to build? · SaaS app (with DB and user auth)
✔ ❯ Select a DB Provider · Sqlite
✔ ❯ Select your background worker type · Async (in-process tokio async tasks)
✔ ❯ Select an asset serving configuration · Client (configures assets for frontend serving)
```

### First thing

1. `$ cargo loco generate model post title:string! content:string! author:references --migration-only`
2. Fix migration
3. `$ cargo loco db migrate`.
4. `$ cargo loco generate controller posts --api`

This works, but the controller it generates is too bare compared to what the scaffold does. Plus it applies migrations in the wrong order.

### Second thing
1. `$ cargo loco db migrate` (to apply the initial users migration)
2. (repeat first thing)

This also works, but despite applying migrations in the correct order, `$ cargo loco db down` **still** removes the users migration first, meaning the timestamp column on migration table does nothing.

### Third thing
1. `$ cargo loco generate scaffold post title:string! content:string! author:references --api`
2. fix migration to point to users
3. rollback migration

This doesn't work because `_entities/post` is invalid. Apparently it can apply a bad migration, but it can't rollback one.

### Fourth thing
Same as above, but also fixing `_entities/post` before rollback. This removes the users table. Meaning if I want to fix the migration I need to delete the entire database since things get undone in the wrong order. Plus it requires me to fix what I assume to be a generated file that shouldn't be edited.

### Fifth thing
1. `$ cargo loco generate scaffold post title:string! content:string! author:references --api`
2. `$ cargo loco db down` in order to fix only the migration and have it generate `_entities/post` correctly.

Doesn't work. Same as before, it can apply the broken migration but not rollback without fixing `_entities/post`. It'd probably remove the users table first too, so it wouldn't work even if it did get past the rollback.

### Sixth thing
1. `$ cargo loco generate scaffold post title:string! content:string! user:references --api`
2. `$ cargo loco generate migration update_posts_rename_user_to_author`
3. write the renaming migration. both up and down.
4. `$ cargo loco db migrate`

This works but it **still** applies/removes migrations in the wrong order. Meaning it's now impossible to rollback these changes since the order in which things will get undone will remove the posts table before the rename gets rolled back. And so it'll reference a non existing table and fail if I try to rollback. Plus it requires me to not only to have multiple migrations for every entity, but to manually write the second one. And if I had a `created_by` and a `updated_by` I'd need to have rename users to `created_by` and add `updated_by`, since generating the model with `user:references user:references` is invalid (this would also require 3+ migrations if I wanted to keep one thing per migration).

**Expected Behavior**

1. Migrations should be applied in the correct order. 
2. Migrations should be rolled back in the correct order. 
3. Generate scaffold shouldn't apply an invalid migration. Or be able to rollback the invalid migration it generated. Or not allow `author:references` if there isn't an authors table and allow multiple references to the same table.

**Environment:**
```
Nome do sistema operacional:                Microsoft Windows 11 Pro
Versão do sistema operacional:              10.0.22631 N/A compilação 22631
rustup:                                     1.27.1 (54dd3d00f 2024-04-24)
rustc:                                      rustc 1.80.1 (3f5fd8dd4 2024-08-06)
cargo:                                      1.80.1 (376290515 2024-07-16)
loco-cli:                                   0.12.0
```

**Additional Context**

Both 1 and 2 (in expected behavior) seem to be fixable by changing the order of the vec in `migration\src\lib.rs`, maybe something like `// inject-above` instead of `// inject-below` or even generating the vec based on the name of the migration file, the migration timestamp is already there and all.
As for 3, something like `created_by:references(users) `updated_by:references(users)` would be great.

ABruel:
Fixes #949 

BREAKING CHANGE: requires inject comment in `migration/src/lib.rs` to now be `inject-above` (and be moved to be after last item)

### DIFF
--- a/examples/demo/Cargo.lock
+++ b/examples/demo/Cargo.lock
@@ -2663,6 +2663,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "tokio-cron-scheduler",
+ "tokio-util",
  "toml",
  "tower 0.4.13",
  "tower-http",
--- a/examples/demo/migration/src/lib.rs
+++ b/examples/demo/migration/src/lib.rs
@@ -2,9 +2,7 @@ pub use sea_orm_migration::prelude::*;
 
 mod m20220101_000001_users;
 mod m20231103_114510_notes;
-
 mod m20240416_071825_roles;
-
 mod m20240416_082115_users_roles;
 pub struct Migrator;
 
@@ -12,11 +10,11 @@ pub struct Migrator;
 impl MigratorTrait for Migrator {
     fn migrations() -> Vec<Box<dyn MigrationTrait>> {
         vec![
-            // inject-below (do not remove this comment)
             Box::new(m20220101_000001_users::Migration),
             Box::new(m20231103_114510_notes::Migration),
             Box::new(m20240416_071825_roles::Migration),
             Box::new(m20240416_082115_users_roles::Migration),
+            // inject-above (do not remove this comment)
         ]
     }
 }
--- a/examples/demo/src/controllers/cache.rs
+++ b/examples/demo/src/controllers/cache.rs
@@ -1,6 +1,7 @@
-use crate::models::users;
 use loco_rs::prelude::*;
 use serde::{Deserialize, Serialize};
+
+use crate::models::users;
 #[derive(Serialize, Deserialize)]
 pub struct CacheResponse {
     value: Option<String>,
--- a/loco-gen/src/templates/migration.t
+++ b/loco-gen/src/templates/migration.t
@@ -6,7 +6,7 @@ skip_glob: "migration/src/*_{{mig_name}}.rs"
 message: "Migration for `{{name}}` added! You can now apply it with `$ cargo loco db migrate`."
 injections:
 - into: "migration/src/lib.rs"
-  after: "inject-below"
+  before: "inject-above"
   content: "            Box::new({{module_name}}::Migration),"
 - into: "migration/src/lib.rs"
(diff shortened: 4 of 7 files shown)


ABruel:
I didn't know if I should change [CHANGELOG.md#39](https://github.com/loco-rs/loco/blob/master/CHANGELOG.md?plain=1#L39) since 0.12 is already out.

jondot:
Looks good. I put back the "below" comment, so that older version of `gen` will still work against `starters` (which are always pointing to master).
In `v0.1.4.0` we're radically changing how starters are created, so that generators and starters are going to be coupled + we're going to have an amazing "new app" generator subsystem that will allow for any kind of app to be generated (using `rhai` scripts)

PR #955, opened 2024-11-03
Named references

### LINKED ISSUE #814: more complex references?
## I can't see how to add a reference that is named

The generate scaffold command seems to support things of the form <type>:reference and <type>:references for letting the type being generated to link to <type>.

In my use-case, I am writing a choose your own adventure style document. So each story_page has a title and some text. Then, each story_page needs to have zero or more links to other pages, each link with a brief description. This is a classic join table, I think, but the relation would look something like:

```
link_from : story_page.id
link_to: story_page.id
```

where this is a many-to-many, but on the story_page view, we want to see all the outbound links, and possibly all inbound links as well for debug during editing.

I can't see how to generate a scaffold for this kind of relationship though. I can do story_page:references which would be fine for one of these two links, but then what do I do for the other one? And how do I name the links by role?

I have probably missed something obvious in the documentation, but I can't find it :/

ABruel:
Add the ability to generate references with custom column names by using `:references:<table>` as a type.

Also fixes generation to allow for multiple references to the same table as in
```
$ generate scaffold posts title:string! content:string! written_by:references:users approved_by:references:users
```

#949 (raised concerns, not the underlying issue)
Fixes #814

### DIFF
--- a/loco-gen/src/templates/model.t
+++ b/loco-gen/src/templates/model.t
@@ -48,7 +48,7 @@ impl MigrationTrait for Migration {
                     {% for ref in references -%}
                     .foreign_key(
                         ForeignKey::create()
-                            .name("fk-{{plural_snake}}-{{ref.0 | plural}}")
+                            .name("fk-{{plural_snake}}-{{ref.1 | plural| snake_case}}")
                             .from({{model}}::Table, {{model}}::{{ref.1 | pascal_case}})
                             .to({{ref.0 | plural | pascal_case}}::Table, {{ref.0 | plural | pascal_case}}::Id)
                             .on_delete(ForeignKeyAction::Cascade)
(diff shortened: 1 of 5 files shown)


loco-gen/src/templates/model.t
```diff
@@ -48,7 +48,7 @@ impl MigrationTrait for Migration {
                     {% for ref in references -%}
                     .foreign_key(
                         ForeignKey::create()
-                            .name("fk-{{plural_snake}}-{{ref.0 | plural}}")
+                            .name("fk-{{plural_snake}}-{{ref.1 | plural| snake_case}}")
```
ABruel, line 51: Use `ref.0` to allow multiple references to the same table, otherwise `scaffold post created_by:references:users updated_by:references:users` would create two `fk-posts-users`.

loco-gen/src/templates/model.t
```diff
@@ -78,8 +78,7 @@ enum {{model}} {
     {% endfor %}
 }
 
-
-{% for ref in references -%}
+{% for ref in references | unique(attribute="0") -%}
```
ABruel, line 81: Avoid creating two enums when referencing a table multiple times.

PR #957, opened 2024-11-03
fix: model conflict false positive

### LINKED ISSUE #956: False positive for conflict on model generation
**Description**
Creating a scaffold `my_test` followed by another scaffold `test` will end in an invalid state since the migration for the second model won't be created.

**To Reproduce**
`$ cargo loco generate scaffold my_test p:string --api`
`$ cargo loco generate scaffold test o:string --api`

**Expected Behavior**

Both should fully apply their respective injections since the migration names are similar, but not the same.

**Environment:**
Windows. Loco 0.12. 

**Additional Context**
Issue being created solely for tracking. Fix already implemented, pr coming shortly.

ABruel:
Fixes #956

### DIFF
--- a/loco-gen/src/templates/model.t
+++ b/loco-gen/src/templates/model.t
@@ -3,7 +3,7 @@
 {% set module_name = "m" ~  mig_ts ~ "_" ~ plural_snake -%}
 {% set model = name | plural | pascal_case -%}
 to: "migration/src/{{module_name}}.rs"
-skip_glob: "migration/src/*_{{plural_snake}}.rs"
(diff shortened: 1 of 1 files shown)


kaplanelad:
@ABruel 
Why did you add `m????????_??????`? 
Can we check the specific name instead of using regex?

ABruel:
> @ABruel Why did you add `m????????_??????`? Can we check the specific name instead of using regex?

The `?` mean any character. I don't know if this system supports regex and not just glob.

Technically it is possible to check for an exact match since the migrator uses the entire file name, but I assumed there was a reason for the glob exclusion that was already there.

PR #962, opened 2024-11-04
Update updated_at with the current timestamp upon saving

uzyn:
Similar to that of Rails.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -343,6 +343,19 @@ pub type {module_pascal} = Entity;
 
 impl ActiveModelBehavior for ActiveModel {{
     // extend activemodel below (keep comment for generators)
+
+    async fn before_save<C>(self, _db: &C, insert: bool) -> std::result::Result<Self, DbErr>
+    where
+        C: ConnectionTrait,
+    {{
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -341,8 +341,22 @@ fn fix_entities() -> AppResult<()> {
...
 
+#[async_trait::async_trait]
 impl ActiveModelBehavior for ActiveModel {{
     // extend activemodel below (keep comment for generators)
+
+    async fn before_save<C>(self, _db: &C, insert: bool) -> std::result::Result<Self, DbErr>
+    where
+        C: ConnectionTrait,
+    {{
+        if !insert && self.updated_at.is_unchanged() {{
```
kaplanelad, line 352: Could you clarify the reasoning behind checking self.updated_at.is_unchanged()? Should we not update the updated_at timestamp for every modified row?

uzyn: My opinion is that it should only be automatically injected if it's missing. If it is already included from controller as part of the data to be updated, it ought to use the newly specified one. Framework should not be overwriting it, especially silently.

PR #964, opened 2024-11-05
Temporarily disable snipdoc check failure to allow CI to pass until a…

kaplanelad:
Temporarily disable snipdoc check failure to allow CI to pass until additional diagnostics are added.


uzyn:
This is addressed at #958

kaplanelad:
For the Snipdoc project, we need to enhance error reporting and add an option to update the store from the code (currently, updates only flow from the store to the code).

For now, I suggest muting failures to avoid constant content re-alignment, which is likely to conflict with upcoming contributions.

I'd love to hear your ideas on how we could improve [Snipdoc](https://crates.io/crates/snipdoc) to make the process easier for contributors.

PR #965, opened 2024-11-05
Fix Cors default_allow_origins() and add allow_credentials

edg-l:
"any" wasnt parsed correctly, the "*" wildcard is recognized by tower


jondot:
@edg-l looks like this triggered a parsing failure -- it says that it prefers "any" rather than "*", so I've reverted back to "any" in a different PR
could you elaborate on how you arrived at this fix?

PR #979, opened 2024-11-09
Add brotli compression as the main static file precompression algorithm

Innominus:
This PR adds brotli precompression to static asset file serving whilst leaving in gzip precompression as a fallback.
Brotli is widely supported in major browsers and provides much better compression.
If the requester does not accept brotli compression, gzip will be served as a fallback.

Also, apologies, my editor removed trailing white spaces from the .md file. Happy to add them back if you want a cleaner PR.

### DIFF
--- a/docs-site/content/docs/the-app/controller.md
+++ b/docs-site/content/docs/the-app/controller.md
@@ -113,7 +113,7 @@ Note that by-design _sharing state between controllers and workers have no meani
 
 ### Shared state in tasks
 
-Tasks don't really have a value for shared state, as they have a similar life as any exec'd binary. The process fires up, boots, creates all resources needed (connects to db, etc.), performs the task logic, and then the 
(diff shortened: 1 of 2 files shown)


Sillyvan:
Nowdays i see most services default to zstd instead. Brotli is amazing but its also a lot more resource heavy.
Maybe zstd would be a better default

Innominus:
> Nowdays i see most services default to zstd instead. Brotli is amazing but its also a lot more resource heavy. Maybe zstd would be a better default

I'm happy to add zstd as well as the first option. I guess it's up to the maintainers and general consensus on what people would prefer. Can update when there's more input.

Innominus:
Making it configurable in future might be the best outcome.

jondot:
hi, we're going to stay with gzip as it's the most universal
but we'll be happy to accept a PR that makes this more configurable or composable.
closing this for now

Innominus:
This doesn't stop anyone from just using gzip, but if the header accepts brotli as an encoding and you have a brotli version of the file it would serve that instead. 
Understandable though if you don't want these methods clogging up the file and prefer it to just be configurable in future, I can understand that. Thanks.

Issue #677, opened 2024-07-27
Updating to version 0.6.2 causes unresponsive server


ghost:
**Description**

Needed to upgrade loco-rs to latest version to resolve clippy error with time version `0.3.34` crate [see here](https://github.com/time-rs/time/issues/681). Now my server is unresponsive:

![image](https://github.com/user-attachments/assets/1189ab82-28f2-47ae-9d49-dcfc9ba7e854)

**To Reproduce**

Run app  with loco-rs version `0.6.2`, run `cargo loco start`, and then ping the server @ http://localhost:5150/_ping.

**Expected Behavior**

A response a from server. Dowgrading loco-rs to `0.6.1` fixes the problem, but re-introduces the time crate error.

![image](https://github.com/user-attachments/assets/64e9b566-b659-4d5d-8448-dbb1fddadc3d)

**Environment:**
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy

Running in default dev container.

**Additional Context**

Hope this is descriptive enough, let me know if you need more info.

kaplanelad:
@ekehoe-tech, is it working when you are running loco outside of the dev container?

ekehoe-tech:
@kaplanelad  Just checked, it does work outside the dev container.

mathmusicoverlord:
> @kaplanelad Just checked, it does work outside the dev container.

Follow up with this account.

kaplanelad:
I'll check the dev container setting. mybe this [pr](https://github.com/loco-rs/loco/pull/627) is related

trevoranderson:
I ran into this as well and I'm nearly certain it's a side-effect of that PR.  When I ran `my_binary start -b 0.0.0.0` in the docker container instead of using the default `localhost` binding it worked, and hypothetically could also be overridden via the config, but I haven't tried that.

There's a tradeoff on ergonomics between using `localhost` as the default which is nice if you are running the binary directly, especially on mac, vs containers where `localhost` isn't accessible from outside.

mathmusicoverlord:
@kaplanelad Could this be set through an environment variable and then the container environment could specify this binding?

trevoranderson:
Should be able to set the binding and port conditionally based on environment via the config file
```
    #[serde(default = "default_binding")]
    pub binding: String,
    /// The port on which the server should listen for incoming connections.
    pub port: i32,
```

[...]

PR #995, opened 2024-11-19
fix: devcontainer interface binding

endrl:
Add interface binding to the config.

Tested

closes: https://github.com/loco-rs/loco/issues/677
Requires: https://github.com/loco-rs/loco-devcontainer/pull/1

### DIFF
--- a/loco-new/base_template/config/development.yaml.t
+++ b/loco-new/base_template/config/development.yaml.t
@@ -16,6 +16,8 @@ logger:
 
 # Web server configuration
 server:
+  # Interface to bind to
+  binding: {{ get_env(name="NODE_BINDING", default="localhost") }}
   # Port on which the server will listen. the server binding is 0.0.0.0:{PORT}
   port: 5150
   # The UI hostname or IP address that mailers will point to.
--- a/loco-new/base_template/config/test.yaml.t
+++ b/loco-new/base_template/config/test.yaml.t
@@ -16,6 +16,8 @@ logger:
 
 # Web server configuration
 server:
+  # Interface to bind to
+  binding: {{ get_env(name="NODE_BINDING", default="localhost") }}
   # Port on which the server will listen. the server binding is 0.0.0.0:{PORT}
   port: 5150
   # The UI hostname or IP address that mailers will point to.


jondot:
@endrl thanks for this PR! also thanks for adjusting and updating the base_template, i'm happy you found it.
this change is hard to accept since `NODE_BINDING` is not a universal thing -- is there a different way to achieve what you're trying to solve?

endrl:
Thank you for you fast response,
i tinkered a little bit and this is what i noticed in docker
```sh
# with interface localhost <-- broken port forward
 sudo netstat -tulpn | grep LISTEN
tcp        0      0 127.0.0.11:39899        0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:45457         0.0.0.0:*               LISTEN      258/node            
tcp6       0      0 ::1:5150                :::*                    LISTEN      1572/target/debug/s 

# with interface 0.0.0.0 <-- works
sudo netstat -tulpn | grep LISTEN
tcp        0      0 0.0.0.0:5150            0.0.0.0:*               LISTEN      1772/target/debug/s 
tcp        0      0 127.0.0.11:46355        0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:38659         0.0.0.0:*               LISTEN      258/node   

# with interface 127.0.0.1 <-- works
sudo netstat -tulpn | grep LISTEN
tcp        0      0 127.0.0.1:5150          0.0.0.0:*               LISTEN      1788/target/debug/s 
tcp        0      0 127.0.0.11:43631        0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:44405         0.0.0.0:*               LISTEN      258/node 
```
A strange ipv6 bind is going on with `localhost`. I don't know why, especially there is no ipv6 network. The node port forward that devcontainer uses is a syscall to `127.0.0.1:PORT`. But `127.0.0.1` is not reachable when loco binds to `localhost`

The vs code docs recommends to bind to `0.0.0.0`.

Possible solutions
- Revert from `localhost` back to `0.0.0.0`
- Change from `localhost` to `127.0.0.1`
- Add a config option, so users can choose. So you can  bind just to a specific ipX network for example. A feature?!

I think the devcontainer setup should work out of the box, i was trapped because of the following config comment 
```
 # Port on which the server will listen. the server binding is 0.0.0.0:{PORT}
```

jondot:
this is solved in upcoming version, we have a new `BINDING` env var.

PR #1042, opened 2024-12-01
Tiny doc improvement: prefer let mut item: ActiveModel = Default::default();

greyblake:
Hi, thanks for the work you do!

I was reading through the docs, and noticed that one thing can be slightly shorten.
Instead of 
```rs
    let mut item = ActiveModel {
        ..Default::default()
    };
```
One could write
```ts
let mut item: ActiveModel = Default::default();
```

The PR also fixes indentation in the `add` function (4 spaces instead of 2 spaces).


docs-site/content/docs/getting-started/guide.md
```diff
@@ -586,7 +584,7 @@ pub fn routes() -> Routes {
         .add("/", post(add))
         .add("/:id", get(get_one))
         .add("/:id", delete(remove))
-        .add("/:id", post(update))
+        .add("/:id", patch(update))
```
greyblake, line 587: The scaffold with `--api` option also uses `patch` for update.

jondot:
Thanks, the reason we use the "spread" variant is that often you want to actually plug a few values right there, and so it is more convenient that we use this.
However I cannot create a convincing argument enough :)

greyblake:
Thank you guys!

Issue #1017, opened 2024-11-25
create `loco db seed --dump [folder]`


jondot:
create `db seed`

* `--dump` takes data from DB to folder (yaml format)
* no flag loads the data from a folder into DB (--from optional)

PR #1046, opened 2024-12-02
seed cli command

kaplanelad:
#1017


src/db.rs
```diff
@@ -475,3 +476,169 @@ async fn create_postgres_database(
...
+        if let Some(ref only_tables) = only_tables {
+            if !only_tables.contains(&table) {
+                tracing::info!(table, "skipping table as it is not in the specified list");
+                continue;
+            }
+        }
+
+        tracing::info!(table, "get table data");
+
+        let data_query = format!("SELECT * FROM {table}");
```
jondot, line 560: try some kind of sql escaping for `table` here

jondot (approved):

Issue #972, opened 2024-11-08
Allow fixtures to execute rust code


DenuxPlays:
## Feature Request

**Is your feature request related to a problem? Please describe.**

Currently I need to parse a CSV file and generate IDs on the go which I cannot do in yaml files because their are static.

**Describe the solution you'd like**

Would be great if we have a new method named like `seed_by_function` that takes a function instead of a path.

**Describe alternatives you've considered**

This can sometimes be moved into the migrations crate but as the migration crate has no access to generator, factories etc. it isn't a viable option for everything.

DenuxPlays:
I am working on a pr for this.

kaplanelad:
moving to discussions
https://github.com/loco-rs/loco/discussions/1072.

PR #1061, opened 2024-12-07
Added ability to seed via rust code

DenuxPlays:
Closes: #972

# TODOs

- [x] create generate command


kaplanelad:
Hey, 
What do you think about the example that I wrote you here, #1062?

Do you still think that the seeder is necessary? 

My feeling is that managing the custom seed logic can be more flexible in the `before_run` than the seeder cli command

DenuxPlays:
> Hey, What do you think about the example that I wrote you here, #1062?
> 
> Do you still think that the seeder is necessary?
> 
> My feeling is that managing the custom seed logic can be more flexible in the `before_run` than the seeder cli command

Yeah could be a solution.
My point is that it gets kind of tedious to implement everything on my own.
Especially if other people want to use a good seeding system they would have to create their own system.

I also thought about customizing the auto-seeder with features like, env based seeding, seeding by tag etc.

kaplanelad:
If the seeder becomes available to you, what different seed functions would you write? I'm still not seeing the need for it, and perhaps I'm missing something.

DenuxPlays:
> If the seeder becomes available to you, what different seed functions would you write? I'm still not seeing the need for it, and perhaps I'm missing something.

Let me create a discussion and summarize my intention maybe this will clear up things.

DenuxPlays:
@kaplanelad 
Discussion: https://github.com/loco-rs/loco/discussions/1072

DenuxPlays:
Closing it for now.
One big PR makes more sense.

Discussion here: https://github.com/loco-rs/loco/discussions/1072

Issue #1060, opened 2024-12-07
`seed` function in `Hook` should get AppContext as an argument.


DenuxPlays:
## Feature Request

Currently the seed function only gets the `DatabaseConnection` but it should get the `AppContext` as an argument.

**Describe the solution you'd like**

Yeah basically as I said.
Change the arguments or extend it.

**Describe alternatives you've considered**

None

**Extras**

I would like to work on this if it gets approved.

DenuxPlays:
Makes more sense together with: https://github.com/loco-rs/loco/issues/972

DenuxPlays:
> can you describe the uae case?

Considering the other ticket (#972) it might be useful to get the AppContext as an argument.
For example I have services that need AppContext as an argument to create them.
Currently it's useless as only files can get used to create fixtures.

In my case I have a Snowflake Generator which i need to use inside the fixtures but the generator requires the AppContext as an argument.

TLDR:
It's only useful with #972 if this ticket/pr gets declined this is to no use.

PR #1063, opened 2024-12-07
Fixtures use AppContext instead of DatabaseConnection

DenuxPlays:
# Breaking Change!!

This is a breaking change due to changing the argument of the `Hooks` trait.

# Refs

Needs: https://github.com/loco-rs/loco/pull/1061

Closes: https://github.com/loco-rs/loco/issues/1060

### DIFF
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -516,8 +516,8 @@ Integrate your seed into the app's Hook implementations by following these steps
 impl Hooks for App {
     // Other implementations...
 
-    async fn seed(db: &DatabaseConnection, base: &Path) -> Result<()> {
-        db::seed::<users::ActiveModel>(db, &base.join("users.yaml").display().to_string()).await?;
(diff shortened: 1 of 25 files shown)


DenuxPlays:
I think the loco-new:ci is broken

It's just looping over the same 15 Tests.

DenuxPlays:
> I think the loco-new:ci is broken
> 
> It's just looping over the same 15 Tests.

GitHub will kill it after 6 hours which will result in a "failure".
So nothing I can do here.

kaplanelad:
@DenuxPlays 
Could you explain the use case? Why is it necessary to include the app context within the seed hook?

DenuxPlays:
> @DenuxPlays Could you explain the use case? Why is it necessary to include the app context within the seed hook?

As described in the issue.
Ref: https://github.com/loco-rs/loco/issues/1060#issuecomment-2536214090

DenuxPlays:
Closing it for now.
One big PR makes more sense.

Discussion here: https://github.com/loco-rs/loco/discussions/1072

PR #1067, opened 2024-12-09
ci: add gen deploy

jondot:


jondot:
this needs to wait for 0.14 because of a breaking change in `validator` upgrade.
building a docker image takes the current-crates version
the docker generator has fixes so needs master.

Issue #613, opened 2024-06-13
Magic link auth


askasp:
You have already made a email verification for password login, is it a great endeavour to have an option for "no-password", that is only magic links?

jondot:
I think it may be a great idea - I believe all the pieces are here for it.

If you want to create a draft of how the developer experience will be, we'll be happy to help you submit a PR for it.

kaplanelad:
- create a new col under user model: magic_token, magic_token_expeired (2 min)
- create post endpoint `atuh_magic_link` and get user email
- Send the magic link in an email
- When click on the link generate jwt token

PR #1085, opened 2024-12-15
magic link

kaplanelad:
#613


loco-new/base_template/src/controllers/auth.rs
```diff
@@ -26,6 +26,11 @@ pub struct ResetParams {
     pub password: String,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
```
jondot, line 29: somehow, only if easy i would add a way to validate an allow list for the emails.
for example, accept only emails from "work domains + gmail". which can avoid people using this to spam with fake emails.
then let the user define the function to validate the requested email (we can leave it open to their judgement)

loco-new/base_template/src/controllers/auth.rs
```diff
@@ -145,6 +150,60 @@ async fn current(auth: auth::JWT, State(ctx): State<AppContext>) -> Result<Respo
...
+    Json(params): Json<MagicLinkParams>,
+) -> Result<Response> {
+    let Ok(user) = users::Model::find_by_email(&ctx.db, &params.email).await else {
+        // we don't want to expose our users email. if the email is invalid we still
+        // returning success to the caller
+        tracing::debug!(email = params.email, "user not found by email");
+        return format::empty_json();
+    };
+
+    let user = user.into_active_model().create_magic_link(&ctx.db).await?;
```
jondot, line 178: a user overwriting a user with `create_magic_link` is a bit confusing
maybe 
let mut user 
user.create_magic_link(..);

because actually it sets the magic link on the user.

jondot (approved):

PR #1086, opened 2024-12-15
add: infer migration

jondot:


DenuxPlays:
Does this adds the ability to generate migrations based on the current db schema?
Like there is a command for generating entities from the current db schema `cargo loco db entities`

jondot:
Not this one but actually it may be a good idea to do separately

PR #1091, opened 2024-12-18
chore: attempt to integrate loco with apalis

geofmureithi:
This PR attempts to combine apalis and loco for background processing. 
The current attempt is for sqlite but it should be easy to apply it to other backends.

Open to a discussion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,6 +143,14 @@ ulid = { version = "1", optional = true }
 rusty-sidekiq = { version = "0.11.0", default-features = false, optional = true }
 bb8 = { version = "0.8.1", optional = true }
 
+# apalis
+apalis = { git = "https://github.com/geofmureithi/apalis.git", branch = "chore/api-improvements", features = [
+    "limit",
+] }
+apalis-sql = { git = "https://github.com/geofmureithi/apalis.git", branch = "chore/api-improvements", features = [
+    "sqlite",
+] }
+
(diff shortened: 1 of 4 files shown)


jondot:
Hi!
This looks interesting and may reduce the amount of custom code we have.
How would it look like if we expand to the other providers such as redis and postgres?
If we can keep the same abstraction in Loco, that could be an easy win

geofmureithi:
> How would it look like if we expand to the other providers such as redis and postgres?

The are all based on the same [Storage](https://docs.rs/apalis-redis/latest/apalis_redis/struct.RedisStorage.html) trait, meaning it should be similar and straight forward.

jondot:
yes this looks good
lets go for it 🚀

jondot:
@geofmureithi any updates?

geofmureithi:
Hey, I have been out for the holidays.
Let me work something between now and the weekend then do a PR on Monday.

geofmureithi:
I am still working on this, the changes will need a new version of apalis, so bear with me as I work through this.

kaplanelad:
@geofmureithi  let me know if you need my help

geofmureithi:
Hey all! I am still working on several things leading to the new minor version release.
These include some needed integration tests and other api changes.
I want to commit to having the v0.7.0-beta and an integration with loco by the end of the month.

PR #1092, opened 2024-12-18
Make internal server error new fallback error type

georgeboot:
Currently, all errors become a 400 Bad Request, unless an explicit error type is set. This is not ideal, as a bad request [should not be retried](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) by a client since it's the client that is at fault (as is for al 4xx codes). A better generic 'fallback' error would be a 500 internal server error, as i[t indicates](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) _something_ went wrong, and sending another attempt might very well succeed.

This pull requests changes the currenty behaviour by returning a 500 on all non-explicit errors.

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -198,10 +198,6 @@ impl IntoResponse for Error {
                 StatusCode::NOT_FOUND,
                 ErrorDetail::new("not_found", "Resource was not found"),
             ),
-            Self::InternalServerError => (
-                StatusCode::INTERNAL_SERVER_ERROR,
-                ErrorDetail::new("internal_server_error", "Internal Server Error"),
-            ),
             Self::Unauthorized(err) => {
                 tracing::warn!(err);
                 (
@@ -221,9 +217,13 @@ impl IntoResponse for Error {
                     ErrorDetail::with_reason("Bad Request"),
                 )
             }
-            _ => (
+            Self::BadRequest(err) => (
                 StatusCode::BAD_REQUEST,
-                ErrorDetail::with_reason("Bad Request"),
+                ErrorDetail::new("Bad Request", &err),
+            ),
+            _ => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                ErrorDetail::new("internal_server_error", "Internal Server Error"),
             ),
         };
 
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -223,7 +223,7 @@ impl IntoResponse for Error {
             ),
             _ => (
                 StatusCode::INTERNAL_SERVER_ERROR,
-                ErrorDetail::new("internal_server_error", "Internal Server Error"),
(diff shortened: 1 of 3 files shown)


src/controller/mod.rs
```diff
@@ -219,11 +219,15 @@ impl IntoResponse for Error {
...
-            _ => (
+            },
+            Self::BadRequest => (
                 StatusCode::BAD_REQUEST,
                 ErrorDetail::with_reason("Bad Request"),
             ),
+            _ => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                ErrorDetail::new("internal_server_error", "Internal Server Error"),
+            ),
```
kaplanelad, line 230: We already have an `InternalServer` enum in this match. Replace the `INTERNAL_SERVER_ERROR` variant with InternalServerError.

georgeboot: Apologies for the oversight. Fixed

kaplanelad:
why did you changed the response?

georgeboot:
> why did you changed the response?

I had to implement the bad request error which carries a reason, and then saw that internal server error had some sort of enum style error message and the same but formatted in the description, so figured it was probably prototyped and never fully implemented, so I did it.

src/controller/mod.rs
```diff
@@ -221,9 +217,13 @@ impl IntoResponse for Error {
...
             }
-            _ => (
+            Self::BadRequest(err) => (
                 StatusCode::BAD_REQUEST,
-                ErrorDetail::with_reason("Bad Request"),
+                ErrorDetail::new("Bad Request", &err),
+            ),
+            _ => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                ErrorDetail::with_reason("Internal Server Error"),
```
kaplanelad, line 226: This is the original response
```
ErrorDetail::new("internal_server_error", "Internal Server Error"),
```

Why did you change it?

georgeboot: I had to implement the bad request error which carries a reason, and then saw that internal server error had some sort of enum style error message and the same but formatted in the description, so figured it was probably prototyped and never fully implemented, so I added the change to this PR. Happy to submit it in another PR if you like.

kaplanelad: The 500 responses should follow this format:
```json
{
    "error":"internal_server_error",
    "description":"Internal Server Error"
}
```

instead of the format you changed it to:
```json
{
    "error":"Internal Server Error"
}
```

georgeboot: But why the specific format though? The error message looks very arbitrary, the description contains the same as the error itself, where other error types don't have that.

kaplanelad: I believe all errors should follow this format, as it provides a clear, standardized error code for API responses.

georgeboot: Cool, will update

PR #1098, opened 2024-12-23
Add anyhow error

anhnmt:
I'm working with a lot of libraries that are using `anyhow` and found adding anyhow error helps a lot


jondot:
Hi, it is generally accepted that libraries should use something like `thiserror` and not `anyhow`
This is why we use `thiserror` exclusively. 
Please use `map_err` to convert your `anyhow` error to one of the options we have.

Issue #1075, opened 2024-12-12
Setting limit_payload to disabled doesn't disable payload limit


BWStearns:
**Description**

https://github.com/loco-rs/loco/discussions/1069

**To Reproduce**

Try this setting in env/development.yaml and try uploading a 4MB file.

```
  middlewares:
    # Limit payload size for incoming requests.
    limit_payload:
      enable: false
```

**Expected Behavior**

Should upload.

**Environment:**

Dev (though presumably any)

**Additional Context**

Seems to be failing back on the default from axum.

BWStearns:
I went to go work on this today and I couldn't reproduce it. Did either of you get it to repro?

PR #1113, opened 2024-12-29
allow disable limit_payload middleware

kaplanelad:
Issue: #1075 

This PR addresses the issue where the Limit Payload middleware is not loaded when it is configured as false in the Loco configuration. In such cases, the application falls back to the default Axum payload limit.

To ensure consistent behavior, the middleware should always be included. If the user wishes to disable the payload size limitation, they can explicitly set the body_limit configuration to


docs-site/content/docs/the-app/controller.md
```diff
@@ -480,19 +476,19 @@ To disable the middleware edit the configuration as follows:
...
+You can customize or disable this behavior through your configuration file.
 
+### Set a custom limit
 ```yaml
 #...
   middlewares:
     limit_payload:
-      enable: true
       body_limit: 5mb
 ```
```
BWStearns, line 489: Maybe add an example of disabling the limit so they don't need to dig/guess for the param?

````
### Remove the limit

_Caution: this can be dangerous because it would allow arbitrarily large files to be uploaded._

```yaml
#...
  middlewares:
    limit_payload:
      enable: false
```
````

kaplanelad: You're absolutely right, this was part of the initial commit. After discussing it with @jondot, we wanted to understand why users might want the option to disable the limits.

We reviewed the middleware implementation and found no performance concerns with keeping the limits enabled. To ensure protection for users, we've decided to set a reasonable default limit, as it was originally. If a user needs a larger limit, they can easily increase it to suit their needs.

The goal is to strike a balance between safety and flexibility.

WDYT?

BWStearns: So, in practice it's probably fine to set a stupidly large limit when you want "unlimited", but stating clearly that it's unlimited is more clear for future readers, vs wondering why 100GB is set as a limit. It's more about communication of intent than a strict requirement I guess.

For my use case I might be uploading kind of arbitrarily large log files, but they're realistically not often going to be more than a couple GB. I am all onboard with sensible defaults, and leaving the safeties on as a default is a good idea, but for some applications it might be nice to be able to switch them off. 

Since `DefaultBodyLimit::disable` is available in Axum I think it makes sense to allow ergonomic access to it via the middleware settings as long as it is flagged as "this is potentially a massive footgun, think before disabling" in the docs.

kaplanelad: @BWStearns i agree,
 reverted to the first implementation

PR #1118, opened 2024-12-30
fix template ci.yaml lost fronted build.

kun1988:
I updated this fix. But I can't run tests. Please check this fix.

This fix is create saas project has client, it run in github actions will fail.


kaplanelad:
The error you encountered:

```
run generate: TemplateEngine(ErrorSystem("copy_template", Template(Error { kind: Msg("Failed to render '__tera_one_off'"), source: Some(Error { kind: Msg("Variable env.RUST_TOOLCHAIN not found in context while rendering '__tera_one_off'"), source: None }) })))
```

occurred because the template tried to evaluate `{{ env.RUST_TOOLCHAIN }}` as a variable but couldn't find it in the context.

Fix it and also the tests.

kaplanelad (approved):

PR #1137, opened 2025-01-04
 Using configuration instead of env enum for the boot application

### LINKED ISSUE #1132: Using configuration instead of env enum for the boot application.
**Description**

this will allow to extend or change the configuration loading behavior

**Expected Behavior**

```rust

pub trait Hooks: Send {

  async fn load_config(environment_name: &str) -> Result<Config> {
      ... default implementation ....
  }
  
  async fn boot(mode: StartMode, config: &config) -> Result<BootResult>;

}

```

Ghost-str:
it resolves #1132

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,19 +13,10 @@ use async_trait::async_trait;
 use axum::Router as AxumRouter;
 
 use crate::{
-    bgworker::{self, Queue},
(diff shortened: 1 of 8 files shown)


kaplanelad:
Can you describe what are you trying to achieve?

kaplanelad:
PR: https://github.com/loco-rs/loco/pull/1143

Issue #1127, opened 2025-01-02
Loco v0.13.3 is not generating the files for the frontend.


pin3da:
**Description**

After following the instructions in https://loco.rs/docs/getting-started/tour/, the `frontend/` directory only contains an empty `frontend/dist` directory. And therefore fails to `npm install`.

**To Reproduce**

1 Start a new project. I did with the following config:

✔ ❯ App name? · myapp
✔ ❯ What would you like to build? · Saas App with client side rendering
✔ ❯ Select a DB Provider · Sqlite
✔ ❯ Select your background worker type · Async (in-process tokio async tasks)

2 `cd frontend/`
3 `npm install`

**Expected Behavior**

It should build. But it fails with `npm error enoent Could not read package.json:` `npm error enoent This is related to npm not being able to find a file.` 

**Environment:**

- `loco --version` -> `loco 0.13.3`
- OS: Fedora release 38 (Thirty Eight)

pin3da:
After some debugging the following change fixed the setup for me:

```
--- a/loco-new/src/generator/executer/filesystem.rs
+++ b/loco-new/src/generator/executer/filesystem.rs
@@ -88,7 +88,7 @@ impl Executer for FileSystem {
     fn create_file(&self, path: &Path, content: String) -> super::Result<PathBuf> {
         let target_path = self.target_dir.join(path);
         if let Some(parent) = path.parent() {
-            fs_extra::dir::create_all(parent, false)?;
+            fs_extra::dir::create_all(self.target_dir.join(parent), false)?;
         }
 
         let span = tracing::info_span!("create_file", target_path = %target_path.display());
```

With the new change, the `frontend` directory appears inside `myapp` and its content is correct (and the `npm build` succeeds).

If this is going in the right direction, I can send a PR with the changes.

Another drive-by comment: Should this use `std::fs::create_dir_all` instead? I'd be happy to add that in my PR if you agree.

Cheers,
Manuel.

kaplanelad:
sure, please open PR with the fix. 
thanks

pin3da:
PR sent: https://github.com/loco-rs/loco/pull/1139

PR #1139, opened 2025-01-04
Fix path inside `create_file`:

pin3da:
This is based on the report in https://github.com/loco-rs/loco/issues/1127

I tested this by running the modified binary:

```
cd loco-new
cargo build -r
```

Then used the newly created binary to create a new app `loco new`

### DIFF
--- a/loco-new/src/generator/executer/filesystem.rs
+++ b/loco-new/src/generator/executer/filesystem.rs
@@ -13,6 +13,12 @@ pub struct FileSystem {
     pub template_engine: generator::template::Template,
 }
 
+impl From<std::io::Error> for super::Error {
+    fn from(err: std::io::Error) -> Self {
+        super::Error::msg(err.to_string())
+    }
+}
+
 impl FileSystem {
     #[must_use]
     pub fn new(from: &Path, to: &Path) -> Self {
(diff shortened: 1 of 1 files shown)


loco-new/src/generator/executer/filesystem.rs
```diff
@@ -13,6 +13,12 @@ pub struct FileSystem {
     pub template_engine: generator::template::Template,
 }
 
+impl From<std::io::Error> for super::Error {
+    fn from(err: std::io::Error) -> Self {
+        super::Error::msg(err.to_string())
+    }
+}
+
```
kaplanelad, line 21: We use the `thiserror` crate for error handling.
But why is it necessary?

loco-new/src/generator/executer/filesystem.rs
```diff
@@ -88,7 +94,7 @@ impl Executer for FileSystem {
     fn create_file(&self, path: &Path, content: String) -> super::Result<PathBuf> {
         let target_path = self.target_dir.join(path);
         if let Some(parent) = path.parent() {
-            fs_extra::dir::create_all(parent, false)?;
+            std::fs::create_dir_all(self.target_dir.join(parent))?;
```
kaplanelad, line 97: please use `fs_extra`

loco-new/src/generator/executer/filesystem.rs
pin3da: This was needed since `std::fs::create_dir_all` returns a different Error.

But will revert that based on the comment below.

loco-new/src/generator/executer/filesystem.rs
pin3da: Done. I asked about this change in #1127 and assumed it was OK from the response.

kaplanelad (approved):

PR #1140, opened 2025-01-04
fix: errors with custom messages are now properly handled

### LINKED ISSUE #1138: Error::Message always results in bad request
**Description**

The `IntoResponse` implementation for `Error` does not cover `Error::Message` and always falls back to the "catch all"
```rs
_ => (
    StatusCode::BAD_REQUEST,
    ErrorDetail::with_reason("Bad Request"),
),
```

**To Reproduce**

Call `.into_response()` on any `Error::Message`

**Expected Behavior**

Properly serialize the custom error message

**Environment:**
any

**Additional Context**
It's not handled here:
https://github.com/loco-rs/loco/blob/master/src/controller/mod.rs#L196-L226

Mettwasser:
Fixes #1138 

Errors with custom messages are now properly handled.
I wasn't sure what status to return with the `Error::Message` variant, but I found `INTERNAL_SERVER_ERROR` fitting.

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -193,6 +193,11 @@ impl IntoResponse for Error {
         }
 
         let public_facing_error = match self {
+            Self::BadRequest(msg) => (StatusCode::BAD_REQUEST, ErrorDetail::with_reason(msg)),
+            Self::Message(msg) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                ErrorDetail::with_reason(msg),
+            ),
             Self::NotFound => (
                 StatusCode::NOT_FOUND,
                 ErrorDetail::new("not_found", "Resource was not found"),


src/controller/mod.rs
```diff
@@ -193,6 +193,11 @@ impl IntoResponse for Error {
         }
 
         let public_facing_error = match self {
+            Self::BadRequest(msg) => (StatusCode::BAD_REQUEST, ErrorDetail::with_reason(msg)),
+            Self::Message(msg) => (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                ErrorDetail::with_reason(msg),
+            ),
```
kaplanelad, line 200: Hi,
Bad requests typically involve common HTTP response codes. I also think it's not ideal to link a message error to an internal server error.

If you'd like to use custom status codes and messages, why not consider using `Error::CustomError`? It provides a clearer way to represent specific errors while maintaining consistency.

To illustrate, I’ve opened a PR that includes comprehensive test coverage and examples of responses:
#1141

Mettwasser: Oh, this is intentional? I thought it's kinda unintuitive to not return the custom message, as you might want to use this error on the front-end. Using inline CustomError(s) really bloats the function, that's why I want to avoid it

kaplanelad: Yes, this is intentional. It’s generally considered a good practice from both a security and user experience perspective. However, it's important to ensure that detailed error information remains accessible to the development or support team for troubleshooting purposes.

Regarding the use of `CustomError`, you can create a wrapper function in your code that maintains the same response behavior.

If you have any function suggestions that we could provide, feel free to share

Mettwasser: I see, good point. Thank you for clarifying!

PR #1147, opened 2025-01-06
Enhance help text on generators command examples 

kaplanelad:


src/cli.rs
```diff
@@ -178,6 +178,21 @@ enum ComponentArg {
...
+    #[command(after_help = format!(
+    "{}  
+  - Generate empty model:
+      $ cargo loco g model posts
+
+  - Generate model with fields:
+      $ cargo loco g model posts title:string! content:text
+
+  - Generate model with references:
+      $ cargo loco g model movies long_title:string added_by:references:users director:references
```
jondot, line 190: `users:references` will generate a reference for users table using a new `user_id` column in movies table.

`users:reference:admin_id` will generate a reference for users table using custom `admin_id` column in movies table.

PR #1151, opened 2025-01-07
fix references in sqlite

jondot:


kaplanelad:
@jondot 
I have pushed [b1f1847](https://github.com/loco-rs/loco/pull/1151/commits/b1f1847780bfbe90cfd38bfd3b152ba797497b77) to test the "add reference" command.

PR #1159, opened 2025-01-08
feat: flatten (de)serialization of custom user claims

### LINKED ISSUE #1115: Change (de)serialization of `UserClaims` (specifically `claims` field)
**Description**

I am building an RBAC system using the built-in JWT support. When using the claims parameter in src/auth/jwt.rs, I encountered unexpected behavior.

Here's the function signature in question:
```rust
loco_rs::auth::jwt::JWT::new().generate_token(&self, secret: &str, expiration: &u64, claims: UserJwtClaims)
``` 
According to [the JWT introduction](https://jwt.io/introduction), claims should correspond to fields within the JWT payload. However, when passing a value to the claims parameter, it serializes as an embedded object under the claims key in the JWT payload instead of merging with the predefined claims. This behavior differs from the expected result of integrating the custom claims directly into the payload.

Here’s the definition of the claims structure from src/auth/jwt.rs:
 ```rust
 /// Represents the claims associated with a user JWT.
#[derive(Debug, Serialize, Deserialize)]
pub struct UserClaims {
    pub pid: String,
    exp: u64,
    pub claims: Option<Value>,
}
```

**To Reproduce**
```rust
#[derive(Debug, Default, Deserialize, Serialize)]
pub struct UserJwtClaims {
    pub roles: Vec<String>,
}

fn main() -> Result<(), Box<dyn Error>> {
    let custom_claims = Some(serde_json::to_value(UserJwtClaims {
        roles: vec!["admin".to_owned(), "non-admin".to_owned()];
    }).unwrap());
    let res = auth::jwt::JWT::new("PqRwLF2rhHe8J22oBeHy").generate_token(&604800, "PID".to_string(), custom_claims)?;
}

```

**Expected Behavior**

I expected the following JWT payload:
```json
{
  "pid": "PID",
  "exp": 1736099338,
  "roles": ["admin", "non-admin"]
}
```
Instead, the actual JWT payload looks like this:
```json
{
  "pid": "PID",
  "exp": 1736099338,
  "claims": {
    "roles": ["admin", "non-admin"]
  }
}
```

**Environment:**

**Additional Context**

I believe this can be resolved with a small adjustment to use the [#[serde(flatten)]](https://serde.rs/attr-flatten.html) attribute in the UserClaims struct:

 ```rust
 /// Represents the claims associated with a user JWT.
#[derive(Debug, Serialize, Deserialize)]
pub struct UserClaims {
    pub pid: String,
    exp: u64,
    #[serde(flatten)]
    pub claims: Option<Value>,
}
```
This change would allow the claims to merge seamlessly with the predefined claims, achieving the expected behavior.

### PR DESCRIPTION
**Pull request by jorgehermo9**
Closes #1115 

Left some doubts as `// TODO:` in the code, please take a look into that comments when reviewing.

CC @Django-Fakkeldij

### CODE REVIEW

File: src/auth/jwt.rs [hunk 1]
```diff
@@ -61,17 +66,18 @@ impl JWT {
...
+    /// use serde_json::Map;
     /// use loco_rs::auth;
     ///
-    /// auth::jwt::JWT::new("PqRwLF2rhHe8J22oBeHy").generate_token(&604800, "PID".to_string(), None);
+    /// auth::jwt::JWT::new("PqRwLF2rhHe8J22oBeHy").generate_token(604800, "PID".to_string(), Map::new());
     /// ```
     pub fn generate_token(
         &self,
-        expiration: &u64,
+        expiration: u64,
```
jorgehermo9, line 76: `u64` implements `Copy` (https://doc.rust-lang.org/std/marker/trait.Copy.html) so there is no need to use references here, we can simply copy it

src/auth/jwt.rs
```diff
@@ -2,23 +2,28 @@
...
     pub pid: String,
     exp: u64,
-    pub claims: Option<Value>,
+    #[serde(default, flatten)]
+    // TODO: should we wrap this in an Option? `Option<Map<String, Value>>`
+    // so we can use `auth::jwt::JWT::new("PqRwLF2rhHe8J22oBeHy").generate_token(&604800, "PID".to_string(), None);
+    // TODO: serde_json::Map or std::collections::HashMap?
+    // TODO: is it ok to use a generic Map<String, Value> here? Or should we let the user specify their desired typed claim and
+    // use generics to serialize/deserialize it?
+    pub claims: Map<String, Value>,
```
jorgehermo9, line 26: `serde_json::Map`? `HashMap`? `BtreeMap`?

src/auth/jwt.rs
```diff
@@ -61,17 +66,18 @@ impl JWT {
     ///
     /// # Example
     /// ```rust
+    /// use serde_json::Map;
```
jorgehermo9, line 69: Should I hide this import in the doctest or not?
```suggestion
    /// #use serde_json::Map;
```

src/auth/jwt.rs
```diff
@@ -119,18 +125,27 @@ mod tests {
...
 
     #[rstest]
-    #[case("valid token", 60, None)]
-    #[case("token expired", 1, None)]
-    #[case("valid token and custom claims", 60, Some(json!({})))]
-    #[tokio::test]
-    async fn can_generate_token(
+    #[case("valid token", 60, json!({}))]
+    #[case("token expired", 1, json!({}))]
+    #[case("valid token and custom string claims", 60, json!({ "custom": "claim",}))]
```
jorgehermo9, line 130: are those cases enough? too much? any specific case that you miss?

src/auth/snapshots/loco_rs__auth__jwt__tests__valid token and custom boolean claims.snap
```diff
@@ -1,6 +1,5 @@
 ---
```
jorgehermo9, line 1: this snapshot file's diff is marked as "renamed file", but actually I deleted the old `valid token and custom claims` snapshot, as it was unreferenced, and added a new `valid token and custom boolean claims` snapshot, such as the other from this PR.

examples/demo/src/models/users.rs
```diff
@@ -216,12 +216,10 @@ impl super::_entities::users::Model {
...
     /// when could not convert user claims to jwt token
-    pub fn generate_jwt(&self, secret: &str, expiration: &u64) -> ModelResult<String> {
-        Ok(jwt::JWT::new(secret).generate_token(
-            expiration,
-            self.pid.to_string(),
-            Some(json!({"Roll": "Administrator"})),
-        )?)
+    pub fn generate_jwt(&self, secret: &str, expiration: u64) -> ModelResult<String> {
+        let mut claims = Map::new();
+        claims.insert("Roll".to_string(), "Administrator".into());
```
jorgehermo9, line 221: it seems that there was a typo in this line before

should I change it to
```suggestion
        claims.insert("Role".to_string(), "Administrator".into());
```?

File: src/auth/jwt.rs [hunk 7]
```diff
@@ -2,23 +2,28 @@
...
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde_json::{Map, Value};
 
 /// Represents the default JWT algorithm used by the [`JWT`] struct.
 const JWT_ALGORITHM: Algorithm = Algorithm::HS512;
 
 /// Represents the claims associated with a user JWT.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
```
jorgehermo9, line 16: added Eq and PartialEq so we can use `assert_eq!` with this struct

src/auth/jwt.rs
Django-Fakkeldij: Only thing I can say is that it won't be a biggie for me :)

src/auth/jwt.rs
Django-Fakkeldij: serde_json::Map for sure! Didn't know that existed.

src/auth/jwt.rs
Django-Fakkeldij: I wouldn't hide it. I think it makes clear to the user what to import to use this feature.

src/auth/jwt.rs
Django-Fakkeldij: Think this is more than enough. You could have a second look at the link below but it doesn't dictate much about the specific form of claims in a JWT.
 
[JWT Spec](https://datatracker.ietf.org/doc/html/rfc7519#section-4)

examples/demo/src/models/users.rs
Django-Fakkeldij: Think so yeah, can't imagine 'roll' was the intended term here haha

Django-Fakkeldij:
Think this looks great! Thank you for taking the time to fix this issue ❤️

jorgehermo9:
Fixed failed CI

src/auth/jwt.rs
jorgehermo9: yep, any JSON value should be valid as the key's value :)

src/auth/jwt.rs
```diff
@@ -2,23 +2,28 @@
...
 pub struct UserClaims {
     pub pid: String,
     exp: u64,
-    pub claims: Option<Value>,
+    #[serde(default, flatten)]
+    // TODO: should we wrap this in an Option? `Option<Map<String, Value>>`
+    // so we can use `auth::jwt::JWT::new("PqRwLF2rhHe8J22oBeHy").generate_token(&604800, "PID".to_string(), None);
+    // TODO: serde_json::Map or std::collections::HashMap?
+    // TODO: is it ok to use a generic Map<String, Value> here? Or should we let the user specify their desired typed claim and
```
kaplanelad, line 24: Can you please remove the TODOs?

src/auth/jwt.rs
kaplanelad, line 16: It looks like `Eq` and `PartialEq` are only needed for testing purposes. To address this, you can use `#[cfg_attr(test, derive(Eq, PartialEq))]` to conditionally derive them only in test builds.

starters/rest-api/src/models/users.rs
```diff
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use chrono::offset::Local;
 use loco_rs::{auth::jwt, hash, prelude::*};
 use serde::{Deserialize, Serialize};
+use serde_json::Map;
```
kaplanelad, line 5: Starters folder are deprecated and move the `loco-new`. 
please revert all the changes under this path

kaplanelad (commented):
Thanks for opening this PR.

src/auth/jwt.rs
jorgehermo9: Yes, no problem. Left the TODOs so I raise my doubts there and the reviewer could comment them. If you are happy with the current implementation, I'll remove all of them

src/auth/jwt.rs
jorgehermo9: Didn't know that! Thank you!

jorgehermo9:
Hi @kaplanelad, thank you for your review. Addressed the changes you suggested.

kaplanelad:
@jorgehermo9 can you fix the CI, please?

jorgehermo9:
Hi @kaplanelad . The `loco-gen-deploy` action is failing (https://github.com/loco-rs/loco/actions/runs/12927511346/job/36052787934?pr=1159) because the app generated tries to use the loco version `0.14.0` instead of the local one. As we changed the public API of the JWT module (and updated the template in `loco-new` in this PR), the app generated should use the local `loco` crate instead of pulling `0.14.0` from the registry, where the JWT API is not updated with these changes.

To reproduce, in the root directory of this branch, run
```
cargo install --path ./loco-new
loco new -n myapp --db sqlite --bg async --assets serverside -a
```
(those are the commands being ran in https://github.com/loco-rs/loco/actions/runs/12927511346/job/36052787934?pr=1159)

and in `myapp/Cargo.toml` you can see this
image

In order to fix that, we should link to the local `loco` dependency, using the `LOCO_DEV_MODE_PATH` env variable as other CI workflows do https://github.com/loco-rs/loco/blob/1787a76746eac62cfee1b30ce4cb4af4be8bc032/.github/workflows/loco-gen-ci.yml#L73

with the following command in local development (being at the repo root directory)  `LOCO_DEV_MODE_PATH=.. loco new -n myapp --db sqlite --bg async --assets serverside -a` the generated `Cargo.toml` file looks like this
image
and the app can be compiled succesfully.

I updated the CI workflow in https://github.com/loco-rs/loco/pull/1159/commits/cac9e5c72174d4605e7986dce08edcff4fb02912 to address this

jorgehermo9:
Now it is failing because the local `loco-rs` crate is not copied in the `dockerfile` https://github.com/loco-rs/loco/actions/runs/12928528426/job/36055996790?pr=1159

How should we proceed here? If the dockerfile build should pull the dependency from crates.io, we have to publish a new `loco-rs` version containing the public api changes.

The other option is to copy the local loco dependency into the docker build context. Tried that in https://github.com/loco-rs/loco/pull/1159/commits/0897e27623498ec427c530a125116d9c675580a1 but I don't know if it's the best workaround... any suggestions? This is not working properly because I'm copying another workspaces in the `myapp` workspace...

jorgehermo9:
Fixed CI. 

The main problem was that the `loco-gen-deploy` was not able to build from a local `loco-rs` dependency as was pulling it from `crates.io`. In order to test this PR, a local `loco-rs` dependency is needed instead, and that dependency must be copied to the docker build context.

https://github.com/loco-rs/loco/pull/1159/commits/0897e27623498ec427c530a125116d9c675580a1 and https://github.com/loco-rs/loco/pull/1159/commits/18de23880d0a78b17917bc38e7d8a5c3272270d3 addressed this limitation

jorgehermo9:
Fixed CI snapshots in https://github.com/loco-rs/loco/pull/1159/commits/13a76d63499ae12b028e8927736df82319464542

There are some unrelated changes from this PR, such as shuttle deployment changes... I think it is because before, the tests were done to the crates.io `loco-rs` `0.14.0` version instead of the local `loco-rs` `0.14.1` version. Updated those snapshots too

loco-gen/tests/templates/snapshots/generate[docker_file_[404_html]_[None]]@deployment.snap
```diff
@@ -7,7 +7,8 @@ FROM rust:1.83.0-slim as builder
 WORKDIR /usr/src/
 
 COPY . .
-
+#  The `loco` root folder should be moved to this dockerfile path so the context can copy it to the image
+RUN mkdir -p /home/runner/work/loco/loco && mv loco/* /home/runner/work/loco/loco && rm -rf loco
```
Comment (jorgehermo9) [inline loco-gen/tests/templates/snapshots/generate[docker_file_[404_html]_[None]]@deployment.snap:11]: Now this template uses the `LOCO_DEV_MODE_PATH` and as the `loco-gen-ci` action sets that variable, these snapshots were generated with `LOCO_DEV_MODE_PATH=/home/runner/work/loco/loco cargo insta test --review --all-features` in local

jorgehermo9:
Hi @kaplanelad 

As you said
> @jorgehermo9 can you fix the CI, please?

I debugged what  was happening (`loco-gen-ci` was pulling the `loco-rs` crate from crates.io instead of using the local modified dependency) and addressed the issue so the CI passed.

Reverted the changes. You can find what I did in commits https://github.com/loco-rs/loco/pull/1159/commits/cac9e5c72174d4605e7986dce08edcff4fb02912, https://github.com/loco-rs/loco/pull/1159/commits/0897e27623498ec427c530a125116d9c675580a1 and https://github.com/loco-rs/loco/pull/1159/commits/18de23880d0a78b17917bc38e7d8a5c3272270d3

kaplanelad (approved):

PR #1169, opened 2025-01-10
fix: bump shuttle to 0.51.0

jonaro00:
- Bumps shuttle deps to latest version
- Updates Shuttle.toml configuration for shuttle.dev platform

Solves https://github.com/shuttle-hq/shuttle/issues/1937

Tested by doing `loco new`, `cargo loco generate deployment` (shuttle), applying the changes in this PR by hand, copying boilerplate from `development.yaml` to `production.yaml`, then `shuttle deploy`. `shuttle run` also works as expected.

### DIFF
--- a/loco-gen/src/templates/deployment/shuttle/shuttle.t
+++ b/loco-gen/src/templates/deployment/shuttle/shuttle.t
@@ -39,7 +39,12 @@ async fn main(
         shuttle_runtime::Environment::Local => Environment::Development,
         shuttle_runtime::Environment::Deployment => Environment::Production,
     };
-    let boot_result = create_app::<App{% if with_db %}, Migrator{% endif %}>(StartMode::ServerOnly, &environment)
+
+     let config = environment
+        .load()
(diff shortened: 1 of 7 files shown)


loco-gen/src/templates/deployment/shuttle/shuttle.t
```diff
@@ -17,7 +17,7 @@ injections:
 - into: Cargo.toml
   after: \[dependencies\]
   content: |
-    shuttle-axum = "{{shuttle_runtime_version}}"
+    shuttle-axum = { version = "{{shuttle_runtime_version}}", default-features = false, features = ["axum-0-7"] }
```
kaplanelad, line 20: can you remove the axum 7 support?

jonaro00:
Another optional improvement would be to add more static asset paths to the Shuttle.toml file, such as `frontend/dist`. This is template dependant, so it could be done preemptively or based on which template is chosen.

PR #1170, opened 2025-01-10
Fix #1119 csr remove unecessary code

### LINKED ISSUE #1119: Selecting `Saas App with client side rendering` creates unecessary code
**Description**

When executing `loco new`and choosing `What would you like to build? · Saas App with client side rendering`, some code only relevant to SSR is created.

**To Reproduce**
1. Run `loco new`
2. Select `❯ What would you like to build? · Saas App with client side rendering`
3. Inspect created project

**Expected Behavior**

Specifically the following items are created which are not necessary:

- `impl Hooks for App` section in `app.rs` contains the initializer for `ViewEngine`, not necessary for CSR
- `initializers/view_engine.rs` is not needed for CSR, the file and corresponding mod declarations and imports can be removed
- `asset` folder contains subfolder with unnecessary items. All of them can be removed.
- `fluent-templates` and `unic-langid` dependency can be removed from `Cargo.toml`

**Environment:**
`loco 0.13.3` and `rustc 1.83.0 (90b35a623 2024-11-26)` stable linux toolchain.

**Additional Context**

Discovered while trying out different CSR UI frameworks

JonasCir:
Fixes #1119

Hi,
this PR removes unnecessary elements from the generated project in case the user selects client-side rendering.

This PR:

- renames assets to rendering method. In my opinion this is more readable and cleaner. The user's choice of the rendering method (i.e., client-side vs. server-side) ultimately decides how assets are served.
- fixes all the mentioned items from #1119, namely:
  -  `ViewEngine` initializer not necessary in `src/app.rs` for CSR
  - `src/initializers/view_engine.rs` is not needed for CSR, the file and corresponding mod declarations and imports can be removed
  -  `asset` folder can be removed for CSR
  - `fluent-templates` and `unic-langid` can be removed from `Cargo.toml` when using CSR

Please note that this changes the CLI output of the tool to use the term "rendering method" instead of "assets" when creating a new project.

I hand tested all the variants by running `loco new` and then inspecting the files in question.

Please let me know if there is anything unclear or if there are any open questions.

 
All the best,
Jonas

---

EDIT: I just noticed that I also need to adjust docs, READMEs, and the website. Will do this after a first review pass that signs of on the CLI changes.


kaplanelad (commented):
Could you please revert the renaming of rendering_method? This field name is used in multiple places beyond the Loco CLI, and it would be better for us to maintain consistency across all references.

Another point, can you test the `setup.rhai` script??

JonasCir:
Hi @kaplanelad, 
thank you for the comment, sure will do (this might still take a couple of days though). 

> Another point, can you test the setup.rhai script??

Can you briefly explain what you mean with that? I am happy to add tests for the rhai extension module I wrote, if this is what you mean?

P.S.: Out of curiosity, where else is the term "assets" used? If absolutely get the consistency requirement here. On the other hand, I think that using the term "assets" to differentiate between SSR and CSR (or doing nothing) is not optimal and has potential to confuse users when reading the documentation (and developers alike). If you are interested in this, we can discuss this separately (in a GH discussion) and I am happy to contribute a PR which fixes this across the project.

JonasCir:
Thanks  @kaplanelad ! Happy to do this :)

kaplanelad:
After receiving no responses, I implemented this fix in #1224. Thanks a lot!

Issue #1182, opened 2025-01-15
Able to configure if Scheduled Task should run on start


DenuxPlays:
## Feature Request

It would be great if the we can say that a task should start when the scheduler starts.

For example in my case I have a task that runs every 24 hours.
But I really wanted to say that the task should run on start and after that every 24 hours.

**Describe the solution you'd like**

Just add another argument to the scheduler like `run_on_start` or something like that.

**Describe alternatives you've considered**

Run the task manually or put it in the before_start hook (but this would also start the task when a worker or the server starts).

DenuxPlays:
If this is a feature you are interested in I can open a PR for this.

kaplanelad:
Sound good. 
You can add it to the under the yaml configuration.
For example see `run_on_start` section:
```yaml
scheduler:
  output: stdout
  jobs:
    write_content:
      shell: true
      run: "echo loco >> ./scheduler.txt"
      schedule: run every 1 second
      output: silent
      tags: ['base', 'infra']

    run_task:
      run: "foo"
      schedule: "at 10:00 am"

    list_if_users:
      run: "user_report"
      shell: true
      schedule: "* 2 * * * *"
      tags: ['base', 'users']
 run_on_start:
  - write_content
  - list_if_users
```

PR #1184, opened 2025-01-15
Added `run_on_start` feature to scheduler

DenuxPlays:
Fixes: #1182

### DIFF
--- a/docs-site/content/docs/processing/scheduler.md
+++ b/docs-site/content/docs/processing/scheduler.md
@@ -56,6 +56,7 @@ scheduler:
     run_task:
       run: "foo"
       schedule: "at 10:00 am"
+      run_on_start: true
 
     list_if_users:
       run: "user_report"
@@ -94,6 +95,7 @@ The scheduler configuration consists of the following elements:
         sec   min   hour   day of month   month   day of week   year
         *     *     *      *              *       *             *
         ```
+  * `run_on_start`: By default, `false`. If set to `true`, the job will also run at the start of the scheduler.
     * `shell`: by default `false` meaning executing the the `run` value as a task. if `true` execute the `run` value as shell command
     * `run`: Cronjob command to run. 
         * `Task:` The task name (with variables e.x `[TASK_NAME] KEY:VAl`. follow [here](@/docs/processing/task.md) to see task arguments ). Note that the `shell` field should be false.
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -219,7 +219,7 @@ pub async fn run_scheduler<H: Hooks>(
         println!("{scheduler}");
         Ok(())
     } else {
-        scheduler.run_on_start().await?;
+        scheduler.run_on_start();
 
         Ok(scheduler.run().await?)
     }
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -219,8 +219,6 @@ pub async fn run_scheduler<H: Hooks>(
         println!("{scheduler}");
         Ok(())
     } else {
-        scheduler.run_on_start();
-
         Ok(scheduler.run().await?)
     }
 }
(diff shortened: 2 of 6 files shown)


DenuxPlays:
Going to fix CI.
Also @kaplanelad is there a reason that in the run function `tokio::signal::ctrl_c().await?;` is used instead of `sched.shutdown_on_ctrl_c();`?

kaplanelad (changes requested):
We can simplify the implementation as follows:

First, we can remove the `run_on_start` function. When starting the scheduler, the `run` function already handles all scheduled jobs, making `run_on_start` is redundant.

The `tokio-cron-scheduler` crate that we are using provides a [new_one_shot](https://github.com/mvniekerk/tokio-cron-scheduler/blob/main/src/job/mod.rs#L397) function. This allows a job to be triggered immediately and ensures it runs only once, while also supporting synchronized execution.

Expected implementation:
1. Call `run`, which manages all scheduler jobs (including async and one-time jobs). This eliminates the need for the `run_on_start` function.
2. In the job logic, check if the current job exists in the run_on_start configuration.
3. If it does, use `new_one_shot` to trigger the job immediately and ensure it only runs once.
4. Retain the `new_async` functionality for asynchronous job handling.



This implementation reduces code duplication, centralizes all job-triggering logic, and ensures the job executor handles all jobs efficiently and easy to test ().

kaplanelad:
> Going to fix CI. Also @kaplanelad is there a reason that in the run function `tokio::signal::ctrl_c().await?;` is used instead of `sched.shutdown_on_ctrl_c();`?

yes

DenuxPlays:
> We can simplify the implementation as follows:
> 
> First, we can remove the `run_on_start` function. When starting the scheduler, the `run` function already handles all scheduled jobs, making `run_on_start` is redundant.
> 
> The `tokio-cron-scheduler` crate that we are using provides a [new_one_shot](https://github.com/mvniekerk/tokio-cron-scheduler/blob/main/src/job/mod.rs#L397) function. This allows a job to be triggered immediately and ensures it runs only once, while also supporting synchronized execution.
> 
> Expected implementation:
> 
> 1. Call `run`, which manages all scheduler jobs (including async and one-time jobs). This eliminates the need for the `run_on_start` function.
> 2. In the job logic, check if the current job exists in the run_on_start configuration.
> 3. If it does, use `new_one_shot` to trigger the job immediately and ensure it only runs once.
> 4. Retain the `new_async` functionality for asynchronous job handling.
> 
> This implementation reduces code duplication, centralizes all job-triggering logic, and ensures the job executor handles all jobs efficiently and easy to test ().

Addressed everything.
Also moved the execute job functionality into its own function to reduce code duplication and to ensure that we use the same logic.

kaplanelad (changes requested):
Thanks for addressing the previous changes!  I’ve just noticed a few more things 

**Documentation:**
Could you please add details about this feature to the [loco.rs documentation](https://loco.rs/docs/processing/scheduler/)?

**CLI:**
Running this command:
```sh
cargo loco scheduler --list
```
should display all scheduler configurations. Would it be possible to add a new column named `run on start` to the output? This column should indicate true or false based on the configuration.


**Validation**
What happens if a task ID that doesn’t exist is added to the run_on_start field? if do nothing, please add validation and make sure that the scheduler not starting to run

DenuxPlays:
For validation:

I would rather print a warning than stopping the scheduler.
This isn't a reason for the scheduler to not work (in most cases a typo or just forgotten to remove).
Would that be enough or do you really want the scheduler to stop with an error?

DenuxPlays:
I've moved the attribute to the job so you can take a look.
If this is unwanted I can revert this commit.

Docs are still missing

kaplanelad (approved):
Thanks, looks good now

DenuxPlays:
This can be merged then Right?

PR #1192, opened 2025-01-20
[WIP] Kamal deployment

yinho999:
- [X] Add `cargo loco generate deployment --kind kamal` command
- [X] Create customize `config/deploy.yml`, `.kamal/secrets` and `Dockerfile` for both sqlite and postgresql + redis.
- [X] Update `development.yaml` to open `0.0.0.0` bindings.
- [x] Override the default `config/deploy.yml`, `.kamal/secrets` and `Dockerfile`.
- [X] Snapshot testing.
- [ ] Documentation of the use case.
- [ ] Blog / Video Guide


loco-gen/src/lib.rs
```diff
@@ -210,6 +214,9 @@ pub enum Component {
         kind: DeploymentKind,
         fallback_file: Option<String>,
         asset_folder: Option<String>,
+        sqlite: bool,
+        postgres: bool,
+        background_queue: bool,
```
kaplanelad, line 219: I believe that none of the parameters (`fallback_file`, `asset_folder`, `host`, `port`) and also the new params are strictly necessary. 
The deployment function should take a reference to &Config (you already have the config in `cli.rs`), and each deployment type should be able to utilize the configuration that works best for it.

loco-gen/src/templates/deployment/kamal/docker.t
```diff
@@ -0,0 +1,47 @@
+to: "Dockerfile"
```
kaplanelad, line 1: Why are we not using the same docker template that we already have? 

I don't think we need to hold two docker templates

kaplanelad (commented):
I realize that this PR is still a work in progress and has been marked as a draft, but please take a look at my comments.

loco-gen/src/templates/deployment/kamal/docker.t
kaplanelad: The question is, why should SeaORM be included in this docker file.

kaplanelad: We are using `sea-orm-cli` for generating entities. Why would you need to generate entities inside the Dockerfile?

yinho999: My current dockerfile for kamal setup is serving the server within the container instead of just compiling, which requires database connection like sqlite or postgres. Therefore SeaOrm is kinda required?

yinho999: Wait sea orm cli is not required for database connection? Okay I will remove it

kaplanelad: Please check if you can use the current docker file and not create a new one

loco-gen/src/lib.rs
yinho999: @kaplanelad I was trying to use &loco-rs::config::Config in loco gen project. However, importing loco-rs within loco-gen will cause cyclic dependency

kaplanelad:
This pr is still relevant?

yinho999:
> This pr is still relevant?

Sorry, I was working on something else for these. I will continue this PR this week. Sorry for the late response

PR #1199, opened 2025-01-20
[BREAKING] Updated validator to `0.20`

DenuxPlays:
With validator 0.20 non `'static` str can be used in `ValidationErrors` because it now uses `Cow<'static, str>`.

A nice thing to have now.

I haven't opened an issue so if this isn't wanted just close it.


DenuxPlays:
Going to fix ci later

DenuxPlays:
Windows test fails because of some type of sscache error:
image

I think the deployment fails because it uses the loco-rs crate that is currently on crates.io which does not yet support validator 0.20 right?

kaplanelad:
working now

kaplanelad:
can you fix the ci?

DenuxPlays:
Yeah later today or this week when I have time

DenuxPlays:
Fixed the fixable ci if the windows test finish sometimes

Issue #1175, opened 2025-01-12
Upgrade broke static file serving


jtwaleson:
**Description**

After upgrading from 0.13.2 to 0.14.0 my static file serving from frontend/dist stopped working. It now shows the fallback "Welcome to Loco!" page.

**To Reproduce**

Create an app in 0.13.2, put your frontend build in dist/frontend. Configure `config/development.yaml` like below. See that it works and your frontend is served. Then upgrade to 0.14.0 and see that you get the default loco fallback page.

```
server:
  port: 8001
  host: http://localhost:8001
  middlewares:
    static:
      enable: true
      must_exist: true
      precompressed: false
      folder:
        uri: "/"
        path: "frontend/dist"
      fallback: "frontend/dist/index.html"

```

**Expected Behavior**

I see the normal frontend app as on 0.13.2.

**Environment:**

n/a

**Additional Context**

I had a quick look at the changes between 0.13.2 and 0.14.0 but could not see any obvious problems. Maybe the different route parsing in axum 0.8.1 has something to do with it?

kaplanelad:
@jtwaleson 
You need to disable the `fallback` middleware. 

```
middleware:
  fallback:
      enable: false
```

jtwaleson:
> @jtwaleson You need to disable the `fallback` middleware.
> 
> ```
> middleware:
>   fallback:
>       enable: false
> ```

Great, that worked! But is this intended? I would expect the fallback to be only shown when the static middleware also can not find any content?

kaplanelad:
In Axum 0.8, change the ability to configure the `nest` route on `/`. However, if you're working with static assets, 
In SPA: it's important to handle the fallback logic within the static asset routes themselves. 
Server-side rendering, us fallback static asset configuration

jtwaleson:
Alright, would you appreciate a documentation update PR? I found it a bit confusing coming from 0.13.2.

kaplanelad:
Sure, would you like to open the PR, or should I take care of it?

PR #1200, opened 2025-01-20
(docs) add explanation for fallback page

jtwaleson:
Added some (hopefully) useful documentation for disabling the fallback page. As of version 0.14.0 the static handler does not take priority over the fallback handler, so the static handler is useless if the fallback is enabled. This relates to issue https://github.com/loco-rs/loco/issues/1175

Please treat this PR as a suggestion and do with it what you will :)

### DIFF
--- a/docs-site/content/blog/frontend-website.md
+++ b/docs-site/content/blog/frontend-website.md
@@ -151,7 +151,16 @@ Now, run the Loco server again and you should see frontend app serving via Loco
 $ cargo loco start
 ```
 
-If you see the default fallback page, you have to disable the fallback middleware.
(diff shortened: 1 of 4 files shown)


docs-site/content/blog/frontend-website.md
```diff
@@ -151,6 +151,8 @@ Now, run the Loco server again and you should see frontend app serving via Loco
 $ cargo loco start
 ```
 
+If you see the default fallback page, you have to disable the fallback middleware.
```
kaplanelad, line 154: I think it's not clear enough, I miss the reason.

PR #1209, opened 2025-01-23
Scaffold v2

kaplanelad:
This PR enhances the scaffold generator and prepares for a shell UI with registration, login, and logout.

- [x] Support Arrays
- [x] Move field definitions from Tera templates to code
- [x] Expand HTML field support
- [x] Better UI


loco-gen/src/template_engine.rs
```diff
@@ -0,0 +1,379 @@
+use std::collections::HashMap;
```
jondot, line 1: if this file is just about helpers for views (scaffold etc) i would call it tera_scaffold_ext.rs
if its general extensions then:
tera_ext.rs

it's ok to name our engine explicitly (tera) just for the purpose of finding exactly what you're looking for "where is this tera builtin coming from? ah, tera_ext.rs"

loco-gen/src/templates/scaffold/html/base.t
```diff
@@ -16,15 +16,56 @@ message: "Base template was added successfully."
...
-  <div id="content">
-    {% raw %}{% block content %}
-    {% endblock content %}{% endraw %}
-  </div>
-
+<body class="min-h-screen bg-background font-sans antialiased">
+    <div class="relative flex min-h-screen flex-col bg-background">
+        <div class="themes-wrapper bg-background">
+            <div style="--sidebar-width:16rem;--sidebar-width-icon:3rem" class="group/sidebar-wrapper flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar">
+                <main class="relative flex min-h-svh flex-1 flex-col bg-background peer-data-[variant=inset]:min-h-[calc(100svh-theme(spacing.4))] md:peer-data-[variant=inset]:m-2 md:peer-data-[state=collapsed]:peer-data-[variant=inset]:ml-2 md:peer-data-[variant=inset]:ml-0 md:peer-data-[variant=inset]:rounded-xl md:peer-data-[variant=inset]:shadow">
```
jondot, line 23: in the context of a "fresh" scaffold file, tailwind classes make it a very unhappy place to edit
can we create custom classes in our own CSS (or maybe in base) and use a single class in the different view files?

docs-site/content/docs/the-app/models.md
```diff
@@ -129,9 +129,9 @@ These fields are ignored if you provide them in your migration command.
 For schema data types, you can use the following mapping to understand the schema:
 
 ```rust
-("uuid", "uuid_uniq"),
-("uuid_col", "uuid_null"),
-("uuid_col!", "uuid"),
+("uuid^", "uuid_uniq"),
```
jondot, line 132: maybe add a special doc about the new array types?

jondot (approved):

loco-gen/src/templates/scaffold/html/base.t
kaplanelad: done

docs-site/content/docs/the-app/models.md
jcorrie: I'm struggling to get array types to work - what's the correct way to scaffold  an array of strings via the cli?

kaplanelad: @jcorrie The array support exists in the master branch and is still not released as part of Loco version.

the syntax is :
[NAME]:array:string

jcorrie: Ah thanks for explaining - much appreciated

sagoez: Hey @kaplanelad, not pushing it but more to have kind of an ETA: when do you think will be the release of the new version of loco with array support?

kaplanelad: We still have a few tasks to complete in this milestone, but we’re aiming to finish it by the end of next week.

In the meantime, you can update your Cargo.toml to use the GitHub version instead of crates.io if you want to work with the latest changes and avoid being blocked.

PR #1212, opened 2025-01-24
add test for all db types

jondot:


kaplanelad:
Moved this changes into #1209

Issue #1037, opened 2024-11-28
Remove generate commands in release mode


DenuxPlays:
## Feature Request

Hey,
when I compile to release mode the end-user doesn't need the generate commands.

**Describe the solution you'd like**

Simply remove generate commands when compiled in release mode.

**Describe alternatives you've considered**

///

PR #1214, opened 2025-01-24
hide `loco generate` in release builds

jondot:
ref: https://github.com/loco-rs/loco/issues/1037


jondot:
also validated no `loco_gen` used at all in final binary
```
nm target/release/myapp-cli | grep loco_gen 
```

Issue #1181, opened 2025-01-15
Rate Limiting


NexVeridian:
[Tower Governor](https://github.com/benwis/tower-governor) some built in ones that might be useful 

[PeerIpKeyExtractor]: this is the default, it uses the peer IP address of the request.
[SmartIpKeyExtractor]: Looks for common IP identification headers usually provided by reverse proxies in order(x-forwarded-for,x-real-ip, forwarded) and falls back to the peer IP address.
[GlobalKeyExtractor]: uses the same key for all incoming requests

for a global extractor we should use tower govern or tower
tower example, from [here](https://github.com/tokio-rs/axum/discussions/987#discussioncomment-2678595):
```rust
use axum::{error_handling::HandleErrorLayer, http::StatusCode, routing::get, BoxError, Router};
use std::{net::SocketAddr, time::Duration};
use tower::{buffer::BufferLayer, limit::RateLimitLayer, ServiceBuilder};

#[tokio::main]
async fn main() {
    let app = Router::new().route("/", get(|| async {})).layer(
        ServiceBuilder::new()
            .layer(HandleErrorLayer::new(|err: BoxError| async move {
                (
                    StatusCode::INTERNAL_SERVER_ERROR,
                    format!("Unhandled error: {}", err),
                )
            }))
            .layer(BufferLayer::new(1024))
            .layer(RateLimitLayer::new(5, Duration::from_secs(1))),
    );

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    axum::Server::bind(&addr)
        .serve(app.into_make_service())
        .await
        .unwrap();
}
```

tower govern also has [custom_key_bearer](https://github.com/benwis/tower-governor/blob/main/examples/src/custom_key_bearer.rs) which could be used for per user rate limiting, instead of by ip, by grabbing the jwt token then calling the database and adding the user id to the custom_key_bearer, but this option has a database call

jondot:
Thanks!
Would you like to create a configuration section for this and add a PR for people to enable rate limiting via configuration?

The middleware stack is here:
https://github.com/loco-rs/loco/blob/master/src/controller/middleware/mod.rs#L72

Middleware configuration here:
https://github.com/loco-rs/loco/blob/master/src/controller/middleware/mod.rs#L171

Example middleware `limit_payload` here:
https://github.com/loco-rs/loco/blob/master/src/controller/middleware/limit_payload.rs

kaplanelad:
@NexVeridian 
[...]

PR #1217, opened 2025-01-24
Cache any value & redis driver for cache

### LINKED ISSUE #1208: Cache values should be anything that can de (de)serialized
Currently, the cache values are always strings. However, it should be relatively easy to store anything that can be serialized.

I am willing to contribute this change, but am announcing my intention so that I can collect additional suggestions.

georgeboot:
Fixes #1208

WIP

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -224,9 +224,9 @@ pub struct Database {
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(tag = "kind")]
 pub enum CacheConfig {
-    /// Redis queue
+    /// In-memory cache
     InMem(InMemCacheConfig),
-    /// Postgres queue
+    /// Redis cache
     Redis(RedisCacheConfig),
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -49,6 +49,7 @@ pub struct Config {
     pub server: Server,
     #[cfg(feature = "with-db")]
     pub database: Database,
+    pub cache: Option<CacheConfig>,
     pub queue: Option<QueueConfig>,
     pub auth: Option<Auth>,
     #[serde(default)]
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -220,6 +221,30 @@ pub struct Database {
     pub dangerously_recreate: bool,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "kind")]
+pub enum CacheConfig {
+    /// Redis queue
```
jondot, line 227: need to align comments here

jondot:
This looks good so far!

kaplanelad:
I believe we should consider using the Redis implementation provided by [Opendal](https://github.com/apache/opendal/tree/main/core/src/services/redis) instead of writing custom logic for the cache function.

Since we're already using Opendal in Loco, this would only require adding a feature flag. Additionally, Opendal supports multiple cache engines, making it easier to integrate other types of engines in the future if needed. 

WDYT?

kaplanelad:
Maybe @Xuanwo can help you with that

georgeboot:
@kaplanelad ive looked at it, and while Opendal supports redis, it only exposes an api to store and retrieve items. Cache also needs expiry, flush and others, so we can't use it.

The other way however, we might consider some re-use. There is for example #1181 that could benefit from atomic increments in redis, something that is currently not yet in our api. Another use case could be distributed locks via redis.

Should we maybe consider making redis its own thing (just like databases are), and re-use that redis in our cache implementation? And perhaps in a future session driver as well. Laravel (rails for PHP) does something similar.

kaplanelad:
From what I can see, Opendal implements expiry (see [here](https://github.com/apache/opendal/blob/main/core/src/services/redis/core.rs#L56C9-L56C12)).

Have you reviewed their implementation?

Reusing your own implementation doesn’t conflict with the idea of leveraging Opendal’s implementation. We can still integrate and reuse their work where applicable.

georgeboot:
@kaplanelad it exists yes, but opendal does not expose it. I don't know why these methods on the redis core are public, as they are all undocumented and you can't create your own redis from from scratch. So doesn't sound like the best idea to use it.

kaplanelad:
I’m not sure I fully understand what you mean by “not expose it.”


If there’s a library we’re already using in other parts of the codebase, why avoid using it? Is there a specific technical issue preventing us from doing so?

The Redis implementation in Opendal addresses use cases that are not covered in your implementation.

georgeboot:
The redis core you refer to basically wraps a redis connection. If i were to use it, the cache adapter would wrap the redis core which wraps the redis connection.

Rails and Laravel make redis its own thing, and reuse one or more redis connections for queues, caches, session, distributed locks, distributed rate limiters etc. Most of them depend on specific functionalities.

Sure, the redis core from opendal support some redis functionalities, but in the near future, when rate limiters or will be implemented, we will find ourselves writing our own redis store so that we have access to the raw redis connection, so we can do things like atomic counters (rate limiting) or atomic sets (distributed locks).

Opendal in the end also uses bb8 and  redis crate, and our implementation is just a thin layer, which will give lots of possibilities in the future.

kaplanelad:
@georgeboot I think it would be a good to split this PR into two parts.

The first PR can focus on making the Cache accept any value, which is what you already did with DeserializeOwned. In the second PR, we can continue our discussion and implement the Redis integration.

What do you think?

Issue #1156, opened 2025-01-08
Remove Example Projects


kaplanelad:
The example projects in the examples folder were initially added to facilitate testing the framework by providing a simple way to test features and fixes.

Over time, we’ve significantly enhanced testing across the framework, as well as in related projects like `rrgen` and `loco-new`, especially when creating new templates. Given these improvements, the example projects are no longer necessary and can be removed.

This cleanup will reduce clutter in the repository while ensuring that our testing infrastructure remains streamlined and focused.

ChadBartley:
were these examples moved somewhere else?

ChadBartley:
No specific need. They were just nice to have for reference while building my project.

I know Rust and the various crates (Axum, SeaORM, Tokio, etc...) all have decent examples. These, being Loco specific were nice to have.

Paulswith:
I think I still need examples, especially when I am not an experienced backend developer.

Paulswith:
> I think I still need examples, especially when I am not an experienced backend developer.

Documentation and examples are very important for a framework (documentation wise, you have done a good job so far). More reference implementations would help lower the learning curve. Did you do any customer surveys before deleting them? (Just curious, no offense)

kaplanelad:
I'm not offended at all—I appreciate feedback from the community! I believe it only helps improve the framework.

No, I didn’t ask before deleting. The Examples project didn’t cover everything the framework offers. Initially, it was created for testing the framework, but once testing was fully integrated into the framework itself, I decided to remove it.

Could you clarify what specific challenges you're facing? Maybe we can find a solution together.

Paulswith:
My bad. I looked for the examples before deleted and found that they were indeed just some test demos, not small projects with targeted implementations like those in axum/example, this is the example I expect... btw, thanks

eibrahim95:
Hi @kaplanelad The [Contributing guide](https://github.com/loco-rs/loco/blob/master/CONTRIBUTING.md) still links to this deleted examples/demo, Can you update it please?

kaplanelad:
Thanks @eibrahim95 for bringing this to our attention. Would you like to open a PR for it?

eibrahim95:
Sure @kaplanelad , but what would be the alternative? I was going through the Contributing guide to see how I can contribute, and I found the example app was removed, so how would a developer adding a feature test out an actual app, using a locally modified loco?

kaplanelad:
[...]

PR #1222, opened 2025-01-26
remove example apps

kaplanelad:
https://github.com/loco-rs/loco/issues/1156


ChadBartley:
Were these examples moved somewhere else?

Issue #1230, opened 2025-01-30
there should be a `From<ModelValidationErrors> for ModelError`, or it should be as easy to manually do so


Denommus:
**Description**

When using `ModelResult<()>`, one would expect to call `active_model.validate()?` and it should just work. Otherwise, it would be expected to do `active_model.validate().map_err(|e| ModelError::ModelValidation { errors: e.into() })?`, or something similar to that. Instead, the closest to the expected behavior is `active_model.validate().map_err(|e| ModelError::DbErr(e.into()))?`, which is not exactly correct, semantically.

**To Reproduce**

Try to use `active_model.validate()?` inside a `ModelResult<()>` function. The compilation fails.

**Expected Behavior**

The program should compile.

PR #1233, opened 2025-01-31
fix: smoother model validations

jondot:


jondot:
https://github.com/loco-rs/loco/issues/1230

Issue #1116, opened 2024-12-30
Unable to nest prefixes/routes


DenuxPlays:
**Description**

Nesting prefixes inside `AppRoutes` is impossible.

**To Reproduce**

```rust
        AppRoutes::with_default_routes() // controller routes below
            .prefix("/api")
            .add_route(controllers::status::non_versioned_routes())
            .prefix("/v1")
            .add_route(controllers::user::routes())
            .add_route(controllers::session::routes())
            .add_route(controllers::openapi::routes())
            .add_route(controllers::status::routes())
```

**Expected Behavior**

The route in `controllers::status::non_versioned_routes()` `/status/version` should be reachable at `/api/status/version`

It is actually only reachable at `/v1/status/version`

**Environment:**

**Additional Context**

Is it possible to nest routes like this at all?
Currently the only way I see is to create one, two Routes (one with /api/v1 and one with /api) and then list all routes there.
This would require to make all route functions public and remove all routes() method from every controller.

DenuxPlays:
Same applies to normal `Routes`

DenuxPlays:
Maybe a `nest()` function just like the one that axum uses would be great for `Routes`.
Like:
```rust
Routes::at("api").nest(other_routes)
````

Or
```rust
Routes::new()
    .nest("api", other_routes)
````

DenuxPlays:
Here is my workaround for now:
```rust
use loco_rs::controller::AppRoutes;
use loco_rs::prelude::Routes;

pub struct ExtendedAppRoutes {
    prefix: Option<String>,
    internal_app_routes: AppRoutes,
}

impl ExtendedAppRoutes {
    pub fn empty() -> Self {
        Self {
            prefix: None,
            internal_app_routes: AppRoutes::empty(),
        }
    }

    pub fn prefix(mut self, prefix: &str) -> Self {
        match self.prefix {
            None => self.prefix = Some(prefix.to_string()),
            Some(mut old_prefix) => {
                old_prefix.push_str(prefix);
                self.prefix = Some(old_prefix.to_string())
            }
        }

        self
    }

    pub fn reset_prefix(mut self) -> Self {
        self.prefix = None;

        self
    }

    pub fn add_route(mut self, mut routes: Routes) -> Self {
        let routes_prefix = {
            if let Some(mut prefix) = self.prefix.clone() {
                let routes_prefix = routes.prefix.clone().unwrap_or("".to_string());

                prefix.push_str(routes_prefix.as_str());
[...]

PR #1241, opened 2025-02-03
Added prefix and route nesting to AppRoutes

DenuxPlays:
replaced snapshot testing with normal asserts:

It was faster to replace snapshots with basic asserts then to look through multiple snapshot files with no relevant content.
Hope thats ok but I could revert it.

Closes: https://github.com/loco-rs/loco/issues/1116

### DIFF
--- a/src/controller/app_routes.rs
+++ b/src/controller/app_routes.rs
@@ -38,7 +38,7 @@ impl fmt::Display for ListRoutes {
         let actions_str = self
             .actions
             .iter()
-            .map(std::string::ToString::to_string)
+            .map(ToString::to_string)
             .collect::<Vec<_>>()
             .join(",");
 
@@ -68,33 +68,31 @@ impl AppRoutes {
 
     #[must_use]
     pub fn collect(&self) -> Vec<ListRoutes> {
-        let base_url_prefix = self
-            .get_prefix()
-            // add a leading slash forcefully. Axum routes must start with a leading slash.
-            // if we have double leading slashes - it will get normalized into a single slash later
-            .map_or("/".to_string(), |url| format!("/{}", url.as_str()));
-
         self.get_routes()
             .iter()
             .flat_map(|controller| {
-                let mut uri_parts = vec![base_url_prefix.clone()];
-                if let Some(prefix) = controller.prefix.as_ref() {
-                    uri_parts.push(prefix.to_string());
-                }
+                let uri_parts = controller
+                    .prefix
+                    .as_ref()
+                    .map_or_else(Vec::new, |prefix| vec![prefix.to_string()]);
+
                 controller.handlers.iter().map(move |handler| {
                     let mut parts = uri_parts.clone();
                     parts.push(handler.uri.to_string());
                     let joined_parts = parts.join("/");
 
                     let normalized = get_normalize_url().replace_all(&joined_parts, "/");
-                    let uri = if normalized == "/" {
+                    let mut uri = if normalized == "/" {
                         normalized.to_string()
                     } else {
-                        normalized.strip_suffix('/').map_or_else(
-                            || normalized.to_string(),
-                            std::string::ToString::to_string,
-                        )
+                        normalized
+                            .strip_suffix('/')
+                            .map_or_else(|| normalized.to_string(), ToString::to_string)
                     };
+
+                    if !uri.starts_with('/') {
+                        uri.insert(0, '/');
+                    }
 
                     ListRoutes {
                         uri,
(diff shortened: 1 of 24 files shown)


src/controller/app_routes.rs
```diff
@@ -132,23 +130,134 @@ impl AppRoutes {
...
+    /// # Example
+    ///
+    /// In the following example, you are adding `api` as a prefix and then nesting `v1` within it:
+    ///
+    /// ```rust
+    /// use loco_rs::controller::AppRoutes;
+    ///
+    /// let app_routes = AppRoutes::with_default_routes()
+    ///     .prefix("api")
+    ///     .nest_prefix("v1");
```
kaplanelad, line 157: Can you provide a bit complete example? 
```rust

# /api/auth/...
# /api/v1/home...
let app_routes = AppRoutes::with_default_routes()
     .prefix("API")
     .add_route(controllers::auth::routes())
     .nest_prefix("v1")
     .add_route(controllers::home::routes())

```

src/controller/app_routes.rs
```diff
@@ -132,23 +130,134 @@ impl AppRoutes {
...
+    }
+
+    /// Set a nested route with a prefix. This route will be added with the specified prefix.
+    /// The prefix will only be applied to the routes given in this function.
+    ///
+    /// # Example
+    ///
+    /// In the following example, you are adding `api` as a prefix and then nesting a route within it:
+    ///
+    /// ```rust, no_run
```
kaplanelad, line 179: can you remove the `no_run`?

src/controller/app_routes.rs
```diff
@@ -132,23 +130,134 @@ impl AppRoutes {
...
+    }
+
+    /// Set multiple nested routes with a prefix. These routes will be added with the specified prefix.
+    /// The prefix will only be applied to the routes given in this function.
+    ///
+    /// # Example
+    ///
+    /// In the following example, you are adding `api` as a prefix and then nesting multiple routes within it:
+    ///
+    /// ```rust, no_run
```
kaplanelad, line 207: can you remove the `no_run`?

src/controller/app_routes.rs
```diff
@@ -253,11 +366,23 @@ mod tests {
...
+            ("/normalizer/multiple-end", vec![Method::TRACE]),
+            ("/multiple1", vec![Method::PUT]),
+            ("/multiple2", vec![Method::OPTIONS]),
+            ("/multiple3", vec![Method::PATCH]),
+        ];
+
+        assert_eq!(routes.len(), expected_routes.len());
+        for (i, route) in routes.iter().enumerate() {
+            assert_eq!(route.uri, expected_routes[i.clone()].0);
+            assert_eq!(route.actions, expected_routes[i].1);
```
kaplanelad, line 385: The test is a bit hard to follow and maintain. It loops over one vector while expecting values from another, making it less intuitive.

Additionally, if the order changes, updating the test becomes cumbersome. I prefer using a snapshot here since it avoids the need to worry about ordering.

src/controller/app_routes.rs
```diff
@@ -206,25 +315,29 @@ impl AppRoutes {
...
 
     #[test]
     fn can_load_app_route_from_default() {
-        for route in AppRoutes::with_default_routes().collect() {
-            assert_debug_snapshot!(
-                format!("[{}]", route.uri.replace('/', "[slash]")),
-                format!("{:?} {}", route.actions, route.uri)
-            );
+        let routes = AppRoutes::with_default_routes().collect();
+        let expected_routes = vec![
```
kaplanelad, line 332: see other comment related to this

src/controller/app_routes.rs
DenuxPlays: Okay I'll revert this change.

src/controller/app_routes.rs
DenuxPlays: No

With your current CI this would fail.
As controllers does not exist in the scope.

kaplanelad: So let's make it work. 
Do you want to share the error?

DenuxPlays: I'll take a break in a few minutes.
I can update the code and commit it then.

DenuxPlays: Sorry didn't had enough time in my break.
We had to tcpdump some shit very cool thing tbh ^^

But once the ci failed you can take a look.
I think it tries to validates the rust code blocks and it will complain that there is no `controller::` available.

DenuxPlays:
I think I fixed the doc test.

I've added the `ignore` attribute.
Not sure if this is what we want as it is not exactly recommended but `compile_fail` also isn't a good fit.

If you want you can take a look [here](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) to see what attributes are available.

kaplanelad:
> I think I fixed the doc test.
> 
> I've added the `ignore` attribute. Not sure if this is what we want as it is not exactly recommended but `compile_fail` isn't a good fit either.
> 
> If you want you can take a look [here](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) to see what attributes are available.

No, this isn’t what we want—this was actually my original comment 🙂.
Let me take a look and push a fix to your PR.

kaplanelad:
The ignore feature has been removed: https://github.com/loco-rs/loco/pull/1241/commits/b4aa3f6c8b8bb73b7db44914d486d7d31eaf92ca.

Also, I noticed that we don’t have documentation on AppRoutes. Could you please add it here?: https://loco.rs/docs/the-app/controller/

DenuxPlays:
Added some docs

kaplanelad (approved):

Issue #1211, opened 2025-01-24
Add flag `--aio` or `--all` to the `start` command


DenuxPlays:
## Feature Request

**Is your feature request related to a problem? Please describe.**

Currently when I want to distribute my application via Docker I need to host two services because it is not possible to run the server, worker and scheduler in one process.

Also someone else asked about this:
https://github.com/loco-rs/loco/issues/1198#issuecomment-2611434834
This is also where I noticed I would benefit from the feature.

IMO the big point is that both the end-user and developer would benefit from this feature.
Easier deployment for the end-user and easier development as everything can be started with one command.

**Describe the solution you'd like**

Currently there is a flag to start the server and worker ( `--server-and-worker`) I would introduce something similar for the scheduler.

But maybe we could replace this flag with something like (`--components=server,worker,scheduler`) to keep it modular?
This could also be done without any breaking changes. Just dont remove the old flag and deprecate it.

The easiest way would of course be a `--all` flag and I do not think there is a use case where you want to start the server and scheduler but not the worker, but maybe I am wrong there.

**Describe alternatives you've considered**

In docker you could run two services managed by supervisord.
This isn't optimal and not very "dockerly" but it would work.

**Contribution**

Again if this is wanted and we can agree to a solution I am open to contribute.

jondot:
I think we can work this according to frequency of use. Let's build some logic foundations:

1. The "correct" way is to use a different process to run and keep running the different aspects: worker, server, scheduler
2. In accordance to our "simplify" mantra, we currently provide: server+worker, but keep the scheduler separate
3. Due to user requests for more simplification we can expand (2) to include the scheduler

I think here being explicit is better than `--all`, because `--all` does not say what is going to be run today, nor what will run in the future (will we have more runner processes? more kinds of workflows?).

(1)
```
cargo loco start --server-and-worker --scheduler
```

To create symmetry we can do

(2)
```
cargo loco start --server --worker --scheduler
```

You can submit a PR for any of (1) or (2).
In case of (2) we'd like to preserve the `--server-and-worker` flag for backwards compat.

Please note that running (or coordinating) the different processes may be tricky (you need to run a main process and then spin off N threads and `join` on them), so apply special attention in the code block that coordinates and run these.

anhnmt:
I really need this feature

kaplanelad:
@anhnmt This will be included in the upcoming release. https://github.com/loco-rs/loco/milestone/25

PR #1247, opened 2025-02-05
allows to run scheduler as part of cargo loco start

kaplanelad:
#1211

After starting this issue, adding another combination when starting the server became overly complex. So, I decided to implement `--all`.


DenuxPlays:
Hey,
thank you for doing this.

Did you see the this comment: https://github.com/loco-rs/loco/issues/1211#issuecomment-2612011260

Not sure if this is what you mean with `dding another combination when starting the server became overly complex`

jondot (approved):

PR #1274, opened 2025-02-23
handle panics in queue worker

jtwaleson:
Users sometimes write code that panics in background job handlers. Right now the entire worker stops handling jobs when this happens. IMHO it makes sense to treat this as a regular error because that makes it easier to use Loco.

Not sure if the maintainers are open to this approach because panics should maybe be handled in the job handler implementations.

If you would like to proceed with this approach, I'll add tests and apply the same to the other worker implementations. If not, I could add some documentation on how to make job handlers panic resistant.


jondot:
yes i think its a good approach, please continue

jtwaleson:
Ok, this should work! I did not add a the for sidekiq, as retrieving the job status does not seem to be supported. I would not know how to elegantly test this. The Async and Sync workers are left untouched, as I assume these are mostly used in development environments and there failing fast seems preferable.

kaplanelad (approved):

PR #1315, opened 2025-03-21
Non-exhaustive SeaORM enums

billy1624:
The `sea_orm::DatabaseBackend` enum is inherently non-exhaustive. It is better to handle it with wildmatch.

Thanks!!

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -337,6 +337,13 @@ async fn has_id_column(
                 "Unsupported database backend: MySQL".to_string(),
             ))
         }
+        #[allow(unreachable_patterns)]
+        _ => {
+            return Err(Error::Message(format!(
+                "Unsupported database backend: {}",
+                stringify!(db_backend)
+            )))
+        }
     };
 
     Ok(result)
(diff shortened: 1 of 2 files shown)


billy1624:
Ping @kaplanelad :P

kaplanelad:
Hey @billy1624,  

Could you share the reasoning behind why we need this change? i not seeing more backend [supported](https://github.com/SeaQL/sea-orm/blob/ef9973b7d1c399c9c09eb9a4c85bbac362998ac2/src/database/db_connection.rs#L57).  

Personally, I'm not a fan of using `wildmatch`. I prefer being explicit rather than implicit. When adding a new type, making it explicit allows the compiler to highlight all the places that require implementation, rather than relying on search-and-replace, which could introduce potential bugs.

billy1624:
Hey @kaplanelad, thanks for the checking this PR!!

I understand your concerns! We will be adding [the `non_exhaustive` attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) to `sea_orm::DatabaseBackend`. Because we're making SeaORM to support more database backends in the near future so we want our downstream crates to be prepared for that.

Does that make sense?

kaplanelad:
Great to see that you're adding support for more database backends!  

Since we (Loco) are sticking to a specific SeaORM version, I prefer explicitly handling new database backends when we update rather than using a wildcard match. Even returning an `Error("not implemented")` would be a better, as it ensures that any newly added backend is explicitly considered.

kaplanelad:
@billy1624 can I close this PR?

billy1624:
Thanks! @kaplanelad Yes, let's close this PR for now :)

PR #1346, opened 2025-04-03
feat: ability to configure foreign_keys pragma for SQLite

ryands17:
This PR allows configuring the `foreign_keys` pragma for SQLite. Some projects work without foreign keys so this would be a nice addition to the configuration

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -221,9 +221,24 @@ pub struct Database {
     #[serde(default)]
     pub dangerously_recreate: bool,
 
-    // sqlite configuration
-    /// set the foreign key pragma to be enabled or disabled
-    pub enable_foreign_keys: Option<bool>,
+    // sqlite run on start commands
+    /// this can be used to confiure PRAGMAs for SQLite where you can pass all values as a string.
+    /// Default values are:
+    ///
+    /// PRAGMA foreign_keys = ON;
+    ///
+    /// PRAGMA journal_mode = WAL;
+    ///
+    /// PRAGMA synchronous = NORMAL;
+    ///
+    /// PRAGMA mmap_size = 134217728;
+    ///
+    /// PRAGMA journal_size_limit = 67108864;
+    ///
+    /// PRAGMA cache_size = 2000;
+    ///
+    /// PRAGMA busy_timeout = 5000;
+    pub run_on_start: Option<String>,
 }
 
 /// Cache configurations for the application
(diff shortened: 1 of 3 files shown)


kaplanelad (commented):
Thank you, @ryands17, for your pull request.

Instead of configuring each `PRAGMA` setting individually, perhaps we could introduce a more general mechanism. For instance, changing the `enable_foreign_keys` setting to `run_on_start` would allow users to configure any PRAGMA settings they desire. This approach would offer greater flexibility and accommodate a wider range of configurations.

WDYT?

ryands17:
@kaplanelad I think that's a better approach! Would this be through the config itself? So instead of maybe a single value, we could add the options specified currently

kaplanelad:
If we add this setup for each property, it will be irrelevant for Postgres. I think a generic configuration would be a better approach.

ryands17:
@kaplanelad Is there a DB specific config that can be added? I can look into that and add these configurable setting there with defaults if not passed in.

kaplanelad:
You can retain the existing configuration; however, if a new configuration is provided, it will override the existing one.

Configuration Example
```yaml
...
# Database Configuration
database:
  # Database connection URI
  uri: {{ get_env(name="DATABASE_URL", default="sqlite://myapp_development.sqlite?mode=rwc") }}

  # Commands to run when the database starts
  run_on_start:
    PRAGMA foreign_keys = ON;
    PRAGMA journal_mode = WAL;
    PRAGMA synchronous = NORMAL;
    PRAGMA mmap_size = 134217728;
    PRAGMA journal_size_limit = 67108864;
    PRAGMA cache_size = 2000;
...
```


Behavior
* If the run_on_start field is defined in the configuration, it will be used as-is.
* If run_on_start is not provided, the application will use the default commands defined in the code.

WDYT?

ryands17:
@kaplanelad Makes sense! I'm guessing this would just be for SQLite so I can update this PR to contain this if that's ok?

kaplanelad:
Yes, sounds good!

ryands17:
@kaplanelad made the changes, wdyt?

src/config.rs
```diff
@@ -218,6 +218,25 @@ pub struct Database {
     /// various things in development.
     #[serde(default)]
     pub dangerously_recreate: bool,
+
+    // sqlite run on start commands
```
kaplanelad, line 222: Ideally, this value should be supported for other cases as well.
So maybe add a comment explaining what it does?
For example: `Execute query after initializing the DB`

kaplanelad:
> @kaplanelad made the changes, wdyt?

Can you check my note?

ryands17:
@kaplanelad Does this look good?

kaplanelad (approved):

PR #1352, opened 2025-04-07
fix db creation

kaplanelad:


src/db.rs
```diff
@@ -195,7 +195,7 @@ pub async fn create(db_uri: &str) -> AppResult<()> {
         Error::string("The specified table name was not found in the given Postgres database URI")
     })?;
 
-    let conn = extract_db_name(db_uri)?.replace(db_uri, "/postgres");
+    let conn = db_uri.replace(db_name, "/postgres");
```
ahwatts, line 198: If, for some reason, the `db_name` is also in the hostname, username, or password part of the `db_uri`, this is going to generate an invalid connection string, right?

kaplanelad: Good catch — that could definitely be an issue. Want to open a PR with a fix?

PR #1358, opened 2025-04-10
Add very permissive CORS setting

### LINKED ISSUE #1357: Allow Insecure `very_permissive` CORS Config
## Feature Request

The app we're developing uses a frontend that is hosted on a separate sub-domain from our Loco-based backend. In some QA environments, these are even hosted on entirely different domains (think auto-generated preview URLs like vercel, cloudflare pages, etc).  

To support session cookie-based auth, we currently use a custom `tower_http::CorsLayer` middleware that applies different CORS policies per environment.  Since we cannot use `Access-Control-Allow-Credentials: true` with `Access-Control-Allow-Origins: *` together, but can utilize [`tower_http::CorsLayer::very_permissive`](https://docs.rs/tower-http/latest/tower_http/cors/struct.CorsLayer.html#method.very_permissive),  it would be great if we could support this the Loco middleware to remove this custom code.

A config option for the CORS middleware to enable the (dangerous) `very_permissive` CORS setting. This setting is quite insecure, so should definitely be identified as such in whatever setting its exposed as.

```
A very permissive configuration:

Credentials allowed.
The method received in Access-Control-Request-Method is sent back as an allowed method.
The origin of the preflight request is sent back as an allowed origin.
The header names received in Access-Control-Request-Headers are sent back as allowed headers.
No headers are currently exposed, but this may change in the future.
```

Keep using our custom middleware, disabling Loco's CORS setting.

mccormickt:
This PR adds a `permissive` configuration option for the `cors` middleware that essentially disables CORS by mirroring the headers sent in the request. This is useful in testing or with ephemeral/non-prod deployments.

https://docs.rs/tower-http/latest/tower_http/cors/struct.CorsLayer.html#method.very_permissive

```
cors:
  enable: true
  permissive: true
  max_age: 3600
```

Open to naming suggestions as well.  Fixes #1357 .

### DIFF
--- a/src/controller/middleware/cors.rs
+++ b/src/controller/middleware/cors.rs
@@ -33,9 +33,13 @@ pub struct Cors {
     pub allow_credentials: bool,
     /// Max age
     pub max_age: Option<u64>,
-    // Vary headers
+    /// Vary headers
     #[serde(default = "default_vary_headers")]
     pub vary: Vec<String>,
+    /// Very permissive policy that allows all origins, headers, and methods and can be used with credentials.
+    /// This policy is not recommended for production.
+    #[serde(default)]
+    pub permissive: bool,
 }
 
 fn default_allow_origins() -> Vec<String> {
@@ -131,11 +135,15 @@ impl Cors {
             cors = cors.vary(list);
         }
 
+        cors = cors.allow_credentials(self.allow_credentials);
+
+        if self.permissive {
+            cors = cors::CorsLayer::very_permissive();
+        }
+
         if let Some(max_age) = self.max_age {
             cors = cors.max_age(Duration::from_secs(max_age));
         }
-
-        cors = cors.allow_credentials(self.allow_credentials);
 
         Ok(cors)
     }
--- a/src/controller/middleware/cors.rs
+++ b/src/controller/middleware/cors.rs
@@ -11,8 +11,9 @@ use axum::Router as AXRouter;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use tower_http::cors::{self, Any};
+use tracing::warn;
 
-use crate::{app::AppContext, controller::middleware::MiddlewareLayer, Result};
+use crate::{app::AppContext, controller::middleware::MiddlewareLayer, Error, Result};
 
 /// CORS middleware configuration
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -39,7 +40,7 @@ pub struct Cors {
     /// Very permissive policy that allows all origins, headers, and methods and can be used with credentials.
     /// This policy is not recommended for production.
     #[serde(default)]
-    pub permissive: bool,
+    pub very_permissive: bool,
 }
 
 fn default_allow_origins() -> Vec<String> {
@@ -92,6 +93,11 @@ impl Cors {
         // look for '< access-control-allow-origin: https://example.com' in response.
         // if it doesn't appear (test with a bogus domain), it is not allowed.
         if self.allow_origins == default_allow_origins() {
+            // Return an error if using a configuration disallowed by CORS
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS/Errors/CORSNotSupportingCredentials
+            if self.allow_credentials {
+                return Err(Error::string("Invalid CORS configuration. `Access-Control-Allow-Credentials: true` cannot be used with `Access-Control-Allow-Origin: *`"));
+            }
             cors = cors.allow_origin(Any);
         } else {
             let mut list = vec![];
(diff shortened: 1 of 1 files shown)


kaplanelad (commented):
Hey @mccormickt,

When configuring CORS in your application, it's crucial to maintain clear and explicit settings to prevent ambiguity. Utilizing a `permissive` setting can inadvertently override existing CORS configurations, leading to uncertainty about which settings are active.

Recommendations

1. **Disable CORS if Not Required:**

   If your application doesn't require CORS in a specific environment, it's advisable to disable the CORS middleware entirely for that environment.

2. **Explicit Configuration:**

   When CORS is necessary, configure it explicitly to align with your application's requirements. This approach ensures that the intended settings are applied without unintended overrides.

In summary, relying on permissive settings can override other CORS configurations, potentially causing issues where the intended settings are not effectively applied. Therefore, it's best to either disable CORS when it's not needed or configure it explicitly to ensure clarity and prevent configuration conflicts.

mccormickt:
>If your application doesn't require CORS in a specific environment, it's advisable to disable the CORS middleware entirely for that environment.
The browser will automatically send CORS preflight requests when sending API calls to another domain, which fail if CORS is disabled on the API server.  So even in this QA environment example, CORS is "required", even if we need to bypass its intended restrictions.

>When CORS is necessary, configure it explicitly to align with your application's requirements. This approach ensures that the intended settings are applied without unintended overrides.

Unfortunately, in this case, CORS cannot be configured with the appropriate options (Allow Credentials + Allow Origins "*"), but is possible with the aforementioned `very_permissive` setting.

>In summary, relying on permissive settings can override other CORS configurations, potentially causing issues where the intended settings are not effectively applied. Therefore, it's best to either disable CORS when it's not needed or configure it explicitly to ensure clarity and prevent configuration conflicts.

No argument here about these points generally, but I think there is middle-ground here. This recommendation is intended to be a non-default, unsafe option for folks that need to send credentials to a backend from a not-previously-known frontend domain.  If it were named `dangerous_very_permissive_enabled` or something to indicate this, maybe that could help?

kaplanelad:
Why isn't it possible to configure AllowCredentials + AllowOrigins("*")?

Maybe in the code, we can check if AllowOrigins is `"*"` and AllowCredentials is enabled, and in that case, build the config using `very_permissive` instead of using the config directly.

Personally, I prefer this approach over requiring permissive to be set explicitly in the configuration.
Would that work?

mccormickt:
> Why isn't it possible to configure AllowCredentials + AllowOrigins("*")?

That's just part of the design of CORS.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS/Errors/CORSNotSupportingCredentials

> Maybe in the code, we can check if AllowOrigins is `"*"` and AllowCredentials is enabled, and in that case, build the config using `very_permissive` instead of using the config directly.
> Personally, I prefer this approach over requiring permissive to be set explicitly in the configuration. Would that work?

I think that would work for my use case, however, I think there is an argument to be made for explicit vs. implicit configuration.  

I think it comes down to answering this question:
If someone configures CORS in this incompatible way, should we, implicitly, use an insecure workaround to make it work?  Or, alternatively, should we provide a clear config option to use this workaround as an escape hatch?

kaplanelad:
From my perspective, if someone configures CORS in an incompatible way, the system should throw an error to alert them.

And if `very_permissive` can already be set in the YAML, what exactly is the problem you're trying to solve?

Maybe we can come up with a solution that better supports users while still respecting their configuration — without turning it into a black box.

mccormickt:
> From my perspective, if someone configures CORS in an incompatible way, the system should throw an error to alert them.
> 
> And if `very_permissive` can already be set in the YAML, what exactly is the problem you're trying to solve?
> 
> Maybe we can come up with a solution that better supports users while still respecting their configuration — without turning it into a black box.

If I follow, it sounds to me like the desired behavior that satisfies the constraints we've identified is something like:
1. Provide an "escape hatch" config option (default is disabled) that can be enabled to facilitate the use of `very_permissive` middleware that effectively bypasses CORS.
2. Provide a warning log for this when enabled to identify that this is unsuitable for production (similar to `pretty_backtrace`) and that CORS is effectively bypassed.
3. When `Access-Control-Allow-Credentials: true` is combined with `Access-Control-Allow-Credentials: *` and `very_permissive` is _not_ enabled, return an error and fail to start.

mccormickt:
>1. Provide an "escape hatch" config option (default is disabled) that can be enabled to facilitate the use of very_permissive middleware that effectively bypasses CORS.
>2. Provide a warning log for this when enabled to identify that this is unsuitable for production (similar to pretty_backtrace) and that CORS is effectively bypassed.
>3. When Access-Control-Allow-Credentials: true is combined with Access-Control-Allow-Credentials: * and very_permissive is not enabled, return an error and fail to start.

@kaplanelad  - I've updated this PR with the above approach. Let me know if this aligns with your thinking or if you'd like to discuss more.

kaplanelad:
This doesn't address the `very_permissive` configuration.
If you want to use `CORS`, you need to understand how to configure it properly. If `very_permissive` isn't allowed in the config, we can still support it—but not by introducing a new config flag. (not creating a new instance of cors with very_permissive)

Also, it might be helpful to add some documentation explaining this

PR #1374, opened 2025-04-17
fix: pass in `LOCO_CONFIG_FOLDER` to scheduler subprocess

### LINKED ISSUE #1372: Scheduler job execution does not respect `LOCO_CONFIG_DIR`
**Description**

When running with args `start --all`, the scheduler executions fail with a `No such file or directory` error.  This seems due to the fact that the scheduler executions pass in environment variables to a shell command to run it and does not include the recently added `LOCO_CONFIG_FOLDER` variable:

https://github.com/loco-rs/loco/blob/23bf71813fd9fa7347df4b5f2402365b8b377a3d/src/scheduler.rs#L198 

**To Reproduce**

Move the loco config directory into some path that isn't `./config` and set `LOCO_CONFIG_FOLDER` to that path and run the scheduler. You will get an error log similar to:
`failed to execute scheduler job in sub process error: No such file or directory (os error 2)`

**Expected Behavior**

The scheduler should respect Loco's standard environment variables like it does with `LOCO_ENV`.

**Environment:**
Containerized Loco app containing only the compiled binary and the proper environment's config, copied to `/opt/` and pointed to using with `LOCO_CONFIG_FOLDER`.

**Additional Context**
N/A

mccormickt:
Fixes #1372

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -3,7 +3,7 @@
 
 use std::{
     collections::HashMap,
-    fmt, io,
+    env, fmt, io,
     path::{Path, PathBuf},
     sync::OnceLock,
     time::{Duration, Instant},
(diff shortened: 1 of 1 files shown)


mccormickt:
Apologies, it looks like this was due to our container deployment not including a shell.

Issue #1344, opened 2025-04-01
Place for initializers and logger setup to store data that is accessible in on_shutdown


masongup-mdsol:
## Feature Request

**Is your feature request related to a problem? Please describe.**

I am working on using loco for an application that needs to integrate with OpenTelemetry tracing. When an app is initialized, it needs to create an OpenTelemetry TracerProvider. This part is handled neatly by the init_logger override. That TracerProvider also needs to be explicitly shut down before the application exits to ensure that all of the traces have been sent to the server. This could also be handled by the on_shutdown override.

However, it is necessary to retain a reference to the TracerProvider somewhere that the on_shutdown function can access it in order for that shut down to be performed - the OpenTelemetry libraries no longer do this for you. [See this also](https://github.com/davidB/tracing-opentelemetry-instrumentation-sdk/pull/185). This is an awkward thing to do with the current set up of these overridable helpers.

What I have done for now in my test app is to create a `static mut OnceLock` in app.rs to hold this value after init_logger returns so that on_shutdown can access it. This is workable, but seems awkward, as it requires `unsafe` in both functions to compile. Can we think of a better way to handle this case?

**Describe the solution you'd like**

A few ideas:

1. We already have a struct `App` that we `impl Hooks` for to do all this, and the `App` struct is defined and owned by the project code. It would be straightforward to store this and any other such data any other particular app needs at this level in that struct. However, this would require an API change for, at least, `init_logger` and `on_shutdown` to accept a reference, possibly with `mut`, to store such data on. Or perhaps an extra trait like `MoreSophisticatedHooks` would be workable?

2. Perhaps the `AppContext` struct could be extended to have some sort of collection for app-specific types used for setup and teardown of the app. This might work similarly to [http Extensions](https://docs.rs/http/1.3.1/http/struct.Extensions.html).

**Describe alternatives you've considered**

The only alternative that I see is the `static mut OnceLock` I described above, which doesn't seem like a great solution. Especially if any app ends up requiring multiple such resources handled in this way.

kaplanelad:
Hey @masongup-mdsol,

### 1. Custom Config Settings  
You can define custom settings in the `yaml` config file using the `custom` section. This allows you to include arbitrary data as part of your application configuration.  
You can see how it’s handled in the code [here](https://github.com/loco-rs/loco/blob/master/src/config.rs#L71).  


### 2. `TracerProvider` Setup  
Instead of overriding `init_logger`, a better approach might be to override the `serve` hook.  
[...]

PR #1376, opened 2025-04-17
Allow storage of arbitrary custom objects in AppContext

masongup-mdsol:
Proposal for how to address #1344.

I tried out writing some code for both my designs 1 and 2 in the linked issue.

Option 1 seems like it would require more extensive adjustments to the Hooks trait than I had initially anticipated, requiring a `create` method to return `Self`, a signature change to at least some of the methods to accept that, plus requiring adjustments to the rest of the Loco code to actually create the struct and pass it around through the internal methods everywhere that calls one of the methods in the trait that takes `self`. This would in turn require every app using Loco to make at least some changes, whether or not they had any use for this feature.

Option 2 seemed neater and more reasonable. I did have to rearrange some of the code in `cli.rs` a little to always have the config before `init_logger` is called, and the signature of `init_logger` changed, but that was inevitable, and should only affect applications currently using custom loggers. I used the http crate's `Extensions` struct for it - not sure if that's ideal, but the http crate is already in the project, and it serves the purpose pretty well.


kaplanelad:
Thanks, @masongup-mdsol!

Could you please elaborate on the `extensions` implementation?  
How do you expect users to interact with or use it?

masongup-mdsol:
My envisioned use of it looks like so:

```rust
use init_tracing_opentelemetry::tracing_subscriber_ext::{init_subscribers, TracingGuard};
pub struct App;
#[async_trait]
impl Hooks for App {
    fn init_logger(context: &loco_rs::app::AppContext, _env: &Environment) -> Result<bool> {
        let guard = init_subscribers().unwrap();
        context.extensions.write().unwrap().insert(Arc::new(guard));
        Ok(true)
    }
    async fn on_shutdown(ctx: &AppContext) {
        let _guard = ctx.extensions.write().unwrap().remove::<TracingGuard>();
    }
}
```

This allows the proper initialization and shutdown of the OpenTelemetry tracing provider.

In this particular case, the necessary shutdown code is in the guard Drop impl, so it probably isn't necessary to retrieve it in the loco `on_shutdown`. It does force the shutdown to happen at the time that loco specifies though, and also demonstrates how to retrieve the object at shutdown time if it is necessary to perform more explicit shutdown operations there.

I can try to put an example somewhere in the docs if you want.

kaplanelad:
Nice work, @masongup-mdsol — thank you for this improvement!

I think it would be great to document this feature, highlighting its benefits and providing a usage example. Would you be interested in contributing the documentation to [loco.rs](https://loco.rs)?

A good place to add it might be here: [https://loco.rs/docs/extras/pluggability](https://loco.rs/docs/extras/pluggability), but I'm open to suggestions if you have a better spot in mind.

masongup-mdsol:
Sure, I can do that. Looking into how and where would be best to include it.

kaplanelad:
thanks

DenuxPlays:
Could it be considered to use a non-poisanable RwLock implementation like from [parking_lot](https://crates.io/crates/parking_lot)?
This would get rid of the error handeling when calling read() or write().

DenuxPlays:
I've built a DI Container in rust and I used [dashmap](https://crates.io/crates/dashmap) for it.
Way better than RwLock etc. maybe this is an option too?

It actually had very good performance.

Another note:
I used this:
```rs
DashMap<(TypeId, String), Box<dyn Any>>
```
With this setup and a few helper method I was able to even safe the same struct multiple times by supplying a "qualifier".

Just throwing in some ideas and things I came across building something similar.

DenuxPlays:
And another thing we could write a simple axum extractor for this to streamline the usage.

I can help with that if needed/wanted.

kaplanelad:
Hey @DenuxPlays,
you're more than welcome to open a PR to this branch with any changes you think should be made.
Thanks!

masongup-mdsol:
IMO, it depends on the desired future direction of Loco. If the project wants to use something along these lines as the official way to manage state for runtime web requests in Loco, then indeed it would be good to use a better hashmap and RwLock impl and add extractors for it. If we want to limit this to just init and shutdown state, then it may be better to stick with the simpler solution with fewer extra dependencies. Either way is fine with me.

DenuxPlays:
Yes totally agree.
Extractors would still be possible here they wouldn't look that different.

But yeah this is something that the maintainers have to decide.

src/app.rs
```diff
@@ -52,6 +56,8 @@ pub struct AppContext {
     pub storage: Arc<Storage>,
     // Cache instance for the application
     pub cache: Arc<cache::Cache>,
+    /// Storage for any helpful app-specific objects
+    pub extensions: Arc<RwLock<Extensions>>,
```
kaplanelad, line 60: Is it possible to access Extensions directly from Axum instead of from the HTTP request?

DenuxPlays: It must implement Clone there is no other way but most of the time would actually have something like `Arc<InnerService>` so the overhead would be cloning an `Arc` which isn't that big.

masongup-mdsol: On the first, I'm not sure what you mean. The Extensions struct comes from the HTTP crate, since it does most of what this needs and is defined there. Nothing from this comes from any particular HTTP request though.

On the second, yes that's true, and I used an `Arc` for it, as @DenuxPlays said, which has cheap clones. I suppose it's on the app author to not put anything in there that is expensive to clone.

DenuxPlays: Thats Fine for me.

For the http issue we could reuse the DashMap Part but instead of (TypeId, String) just use TypeId for the key.

In this way we would be Independent from the http crate.

Another Option would be to just copy the Extensions Code from the http crate if the license allows it.

kaplanelad: @DenuxPlays Sounds good! I'm working on it right now and will open a PR soon. It would be great to get your code review!

DenuxPlays: Oh I updated my pr to reflect this.
Also simplified it a lot as the whole service can be moved in another crate.

You can take a look and maybe use it as a base or something

DenuxPlays: Finished my pr by adding docs.

However there is another thing I think we forgot.
I do like the change that `init_logger` now has access to `AppContext` but I think we should do that in a separate pr for cleaner Changelogs and git history.

kaplanelad:
@masongup-mdsol Following PR #1404, could you open a PR to add the Context to the logger within the init_logger hook?

Also, please remember to document the breaking changes [here](https://github.com/loco-rs/loco/blob/master/docs-site/content/docs/extras/upgrades.md?plain=1).

PR #1396, opened 2025-04-28
optimize worker tag filtering string handling

kaplanelad:


jtwaleson:
Excellent, thanks a lot @kaplanelad !

PR #1398, opened 2025-04-28
Added Di Container

DenuxPlays:
This is my approach to https://github.com/loco-rs/loco/pull/1376

It is independent from the `http` crate while still achieving good performance by utilizing `DashMap`.
I've also added support for an Axum extractor.

TODO:

- [x] update `Changelog.md`
- [x] add docs section

<details>
<summary>Old</summary>

This is my approach to https://github.com/loco-rs/loco/pull/1376

This would allow the add multiple instances of the same type.

Also I plan to add a LazyService trait which allows for lazy initialization.
Will do that later today.

TODOs:

- [x] add a Data and Service system
- [x] add documentation
- [x] add tests
- [ ] add custom error? (do we want this)
- [ ]  ~add the ability to get a ref~ No I think using Arc<> is the easier and better approach
    - had some issues with that
    - maybe return the `Ref` type provided by DashMap?
    - Returning plain `Option<&T>` is not possible
    - current drawback: user has manually wrap the service in `Arc` if its not clonable (not a huge deal imo)
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Adding cache redis driver + configuration instead of enabling from code. [https://github.com/loco-rs/loco/pull/1389](https://github.com/loco-rs/loco/pull/1389)
 * Ability to configure pragma for SQLite. [https://github.com/loco-rs/loco/pull/1346](https://github.com/loco-rs/loco/pull/1346)
 * Optimize worker tag filtering string handling. [https://github.com/loco-rs/loco/pull/1396](https://github.com/loco-rs/loco/pull/1396)
+* Added DiContainer to manage instances via the `AppContext`. [https://github.com/loco-rs/loco/pull/1398](https://github.com/loco-rs/loco/pull/1398)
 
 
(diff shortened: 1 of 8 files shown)


DenuxPlays:
If this is something you would want in loco then I would also update the documentation on loco.rs

kaplanelad:
Thanks, @DenuxPlays!

I'm not sure I understand—why not just use [settings](https://github.com/loco-rs/loco/blob/master/src/config.rs#L73) directly?

DenuxPlays:
How?
Settings safes some value you define in the config.

A Di Container manages instances of structs.

You cannot use settings.
If you mean the qualifier it you can do stuff like this:

Imagine you have multiple http clients (one for api1 and one for api2) you can differentiate them by inserting them with a different qualifier but most of the time you wouldn't actually use a qualifier it's just for special cases like this.
But I think they occur often enough to properly support them.

kaplanelad:
Share a comment in #1404. It would be great to get your feedback!

PR #1411, opened 2025-05-03
make "self" available to worker tags function

jtwaleson:
This way we can use runtime config changes for which worker to execute this task on. E.g. for development I want to use a single worker, but in production I want multiple separate ones.

### DIFF
--- a/src/bgworker/mod.rs
+++ b/src/bgworker/mod.rs
@@ -600,7 +600,7 @@ pub trait BackgroundWorker<A: Send + Sync + serde::Serialize + 'static>: Send + 
     /// Specifies tags associated with this worker. Workers might only process jobs
     /// matching specific tags during startup.
     #[must_use]
-    fn tags() -> Vec<String> {
+    fn tags(&self) -> Vec<String> {
         Vec::new()
     }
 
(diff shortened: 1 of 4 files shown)


src/bgworker/mod.rs
```diff
@@ -622,7 +622,8 @@ pub trait BackgroundWorker<A: Send + Sync + serde::Serialize + 'static>: Send +
         match &ctx.config.workers.mode {
             WorkerMode::BackgroundQueue => {
                 if let Some(p) = &ctx.queue_provider {
-                    let tags = Self::tags();
+                    let worker = Self::build(ctx);
```
kaplanelad, line 625: Building the worker every time perform_later is called can be problematic, especially if the user has custom logic inside the build function. It's likely that users won't expect the build function to be invoked repeatedly.

jtwaleson: Hm you're probably right. Would it make sense to call `Self::tags(ctx)` instead?

kaplanelad: I don't think so.
Why not add an all tag and run the worker with that tag during development?

jtwaleson: That's an interesting idea. I think that for development, most people will use `--server-and-worker`. (At least I do.) And when you use that, it doesn't run any task with tags.

kaplanelad: Tags are not currently supported when using the --server-and-worker flag.
You can define a tags function like this:

```rust
fn tags() -> Vec<String> {
    if cfg!(debug_assertions) {
        vec![]
    } else {
        vec!["cpu".to_string()]
    }
}

```
Alternatively, you can use compile-time configuration with #[cfg] attributes:
```rust
fn tags() -> Vec<String> {
    #[cfg(debug_assertions)]
    {
        return vec![];
    }

    #[cfg(not(debug_assertions))]
    {
        return vec!["cpu".to_string()];
    }
}
```

jtwaleson: Ok, happy to close this PR. I can work with compile time differentiation for now, thanks!

PR #1418, opened 2025-05-05
Update the init_logger to use app_context instead of config

masongup-mdsol:
Per discussion in #1404 and #1344, update the init_logger signature in the Hooks trait to accept AppContext instead of Config to allow access to the recently-added SharedStore.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -21,7 +21,7 @@ use crate::{
     bgworker::{self, Queue},
     boot::{shutdown_signal, BootResult, ServeParams, StartMode},
     cache::{self},
-    config::{self, Config},
+    config::Config,
     controller::{
         middleware::{self, MiddlewareLayer},
         AppRoutes,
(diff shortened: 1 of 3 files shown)


kaplanelad:
Thanks, @masongup-mdsol!

- Could you please document the breaking changes in [upgrades.md](https://github.com/loco-rs/loco/blob/master/docs-site/content/docs/extras/upgrades.md#upgrade-from-015x-to-016x)?
- Have you tried implementing `opentelemetry` to see if it works as expected?

src/app.rs
```diff
@@ -359,7 +359,7 @@ pub trait Hooks: Send {
     ///
     /// # Errors
     /// If fails returns an error
-    fn init_logger(_config: &config::Config, _env: &Environment) -> Result<bool> {
+    fn init_logger(_ctx: &AppContext, _env: &Environment) -> Result<bool> {
```
kaplanelad, line 362: I believe the `_env: &Environment` parameter is redundant.

Since `Environment` is already included as part of `AppContext` (accessible via `ctx.environment`), it's better to pass only the `AppContext` and remove the separate `Environment` parameter.

masongup-mdsol: Good point, done

kaplanelad (approved):

jtwaleson:
@masongup-mdsol do you maybe have an example for the `TracingGuard` implementation? I've configured Sentry with a static variable and would like to switch to something better. I can open a separate PR with some documentation later.

Issue #1402, opened 2025-04-30
Make `build` function async and return Result


DenuxPlays:
## Feature Request

For worker make the `build` function async and return `Result<Self, Error>`.
For tasks create a build function that is async and returns `Result<Self, Error>`.

**Is your feature request related to a problem? Please describe.**

A lot of my workers need some kind of dependencies (managers, api client, snowflake generators etc.) and currently I have to create or get them for every perform of the worker.
If the `build` function would be async and retrun `Result<Self, Error>` dependencies could be initialized on build instead of run.

**Describe the solution you'd like**

Currently I use [block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on) and unwrap() to get around it or create it for every job the worker executes.

Not an ideal solution especially when you have APIs that requires you to send heartbeats these dependencies have to be stored in some kind of global `OnceLock` or something like that.

**Describe alternatives you've considered**

Not really found one but I am open to suggestions.

**Notes**

Again if you agree with me I am open to add and maintain this.

kaplanelad:
Yes you can add `async` for `build` worker.
few notes:
* handle also in [generator](https://github.com/loco-rs/loco/blob/master/loco-gen/src/templates/worker/test.t)
* Adding braking changes not [here](https://github.com/loco-rs/loco/blob/master/docs-site/content/docs/extras/upgrades.md?plain=1)

thanks

DenuxPlays:
Hey @kaplanelad what do you think of abstracting the `build` function into a trait like this:
```rust
pub trait Buildable {
    fn build(ctx: &AppContext) -> impl Future<Output=Result<Self, Error>
}
```

And then the worker trait would look something like this:

```rust
pub trait BackgroundWorker: Buildable {
///
}
```` 

This way we could ensure a define and consistent way between workers, tasks and maybe more in the future?

kaplanelad:
Do you want to try and let me know?

DenuxPlays:
@kaplanelad I've noticed that when we build workers everytime when not working with queues.

With the new storage we could build them one time and then reuse that instance.
I think that is what most (including me) would expect how it works.

Do you agree and should I open a pr for it?

DenuxPlays:
Maybe if we merge everything together we could get rid of the build function (just like we have with `Task`) and instead of `connect_workers` we have `register_workers` and we safe them for later reuse (either directly or by registering them to the queue).

This would inline the workers with the tasks.

[...]

PR #1420, opened 2025-05-05
Enhanced worker system

DenuxPlays:
- User is now in charge on how Workers are created (just like with Tasks)
- Instances of workers are reused instead of created (like someone would expect)
- Now every worker has to be connected

Closes: #1402 

@kaplanelad This is what I had in mind.

This is just a base concept there are still missing templates, docs migration guides etc.

Let me know what you think about it.


kaplanelad:
Hey @DenuxPlays,
Just to confirm—does this PR include making the worker a singleton?

DenuxPlays:
Yes
IMO this is what I thought is the case atm (it is but just when using queues).
So it would do the same regardless if you are using queues or other backends.

kaplanelad:
I'm a bit hesitant about moving the workers to be singletons by default. The current worker design is intentionally stateless—this allows them to be lightweight, easy to scale, and resilient to crashes or restarts.

Making them stateful by turning them into singletons introduces tight coupling and implicit state, which can lead to issues. For example, if a worker holds internal state and the server process is restarted (due to a crash, deployment, or auto-scaling), that state is lost. This could cause inconsistent behavior or even dropped background jobs if not carefully managed.

Another scenario: if the application is running across multiple processes (or containers), a singleton pattern only ensures a single instance per process, not across the whole system. This can result in unexpected concurrency bugs or duplicated work if coordination isn’t handled externally.

Overall, stateless workers offer better flexibility and reliability, especially in distributed environments. If there's a need for stateful behavior in specific use cases, maybe we could offer that as an opt-in rather than the default?

DenuxPlays:
But then we need to figure out a solution for the queue system because it is currently statefull.

I agree with all of that but tbh this is what every other framework I've ever used done it (including big ones like symfony and spring-boot) so it works in production.

kaplanelad:
If you take `ruby sidekiq` for an example, it is a stateless.

DenuxPlays:
Okay haven't used it.

But since loco is based on ruby we should probably go with stateless (statefull workers can be integrated later or via initializers).
Then we need to figure out how to handle queues as their are currently statefull

DenuxPlays:
I am not sure what you mean

kaplanelad:
In the new version, we've introduced a shared_store under AppContext. This allows you to insert any struct into the AppContext, making it accessible within your Jobs.

kaplanelad:
Have you considered using `once_cell` or `lazy_static` for this?

DenuxPlays:
Not an option as this is untestable

Thats why a container is always used

PR #1421, opened 2025-05-06
site: docs: models: add info about `!` & `^`

Frankkkkk:
Thanks!

### DIFF
--- a/docs-site/content/docs/extras/upgrades.md
+++ b/docs-site/content/docs/extras/upgrades.md
@@ -35,6 +35,19 @@ These are the major ones:
 
 ## Upgrade from 0.15.x to 0.16.x
 
+### Use `AppContext` instead of `Config` in `init_logger` in the `Hooks` trait
+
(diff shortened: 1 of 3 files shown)


docs-site/content/docs/the-app/models.md
```diff
@@ -126,6 +126,15 @@ When a model is added via migration, the following default fields are provided:
 
 These fields are ignored if you provide them in your migration command.
 
+### Field syntax
+
+Each field generally follows the format: `name:type[!^]`, where:
```
kaplanelad, line 131: I think you should also include the giver when `!` or `^` is not provided.

Frankkkkk: You mean specify that they are optional?

kaplanelad: Yes we have 3 option:
1. name:string!
2. name:string^
3. name:string

PR #1460, opened 2025-05-27
Commit Cargo.lock

frisoft:
I noticed that Loco does not have Cargo.lock in the repo.

This causes issues when trying to install Loco from sources in Nix. Not having a Cargo.lock makes the installation non-reproducible.

Is there any particular reason for not committing the Cargo.lock file?


DenuxPlays:
I agree with kaplanelad but maybe we can commit it and add it to the `exclude`  field in to Cargo.toml

Not sure if that really works just an idea I wanted to share

frisoft:
My point is that there is no way to have the loco binary built in a reproducible way (same version is the same executable in 2 machines with the same CPU architecture) to run loco new without a common Cargo.lock.

kaplanelad:
You are talking on `loco new` or `loco-rs`?
[loco new](https://github.com/loco-rs/loco/tree/master/loco-new) is a not part of the loco-rs lib and it only a CLI. the [here](https://github.com/loco-rs/loco/tree/master/loco-new) we can add the lock file

PR #1464, opened 2025-05-30
Redis atomic job queue

kaplanelad:


DenuxPlays:
Hey,

Haven't noticed anything but I am not done testing

Just wanted to say that we should probably Release an RC asking more to Test their application to cover Some edge cases we probably don't think about

kaplanelad:
Thanks, @DenuxPlays! Looking forward to your response once you're done testing.

DenuxPlays:
Hey,

sorry for the delay but I've tested it.

I hadn't had any issues with atomcity so this approach should work in production (I would still release an rc to widen our test cases).
I've noticed an increased redis CPU usage compared to the old implementation but nothing drastically. It was so minimal but it may also be just some test fluctuations. (less then 0.2% but consistent)

jtwaleson:
@kaplanelad hope you're well. Let me know if you need any more testing. I doubt that I can do better than @DenuxPlays .

DenuxPlays:
Would be great if you could test it too

If we can cover  more use cases we can release it with more confident or maybe find a bug we can fix before we release it

PR #1468, opened 2025-06-05
Bump version 0.16

kaplanelad:


.github/workflows/loco-gen-ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-features -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms
+          args: --all-features -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms -A clippy::result_large_err
```
DenuxPlays, line 46: Should we really do that?

This lint was enabled for a (performance) reason.

I know that this is nothing big but this could be a chance to "fix"/address it instead of ignoring it.

Also instead of changing up the clippy command we should disable them inside the Cargo.toml `lint` section so that `cargo clippy` just works out of the box.

PR #1478, opened 2025-06-11
Add a launcher config

therustmonk:
Ability to hide the banner via configuration (to save visual space in the terminal)


kaplanelad:
Thanks for the suggestion! I’d prefer not to add this to the YAML config, since it’s more of a developer or runtime concern rather than something that should live in the app’s persistent configuration.

Instead, I’d suggest supporting an environment variable, like:
```
LOCO_NO_BANNER=false
```

kaplanelad:
Closing this PR due to inactivity. Feel free to reopen it if you'd like to continue.
Thanks for your contribution!

PR #1484, opened 2025-06-16
Add CSRF protection by default

floscodes:
Added axum_csrf crate and implement it's csrf middleware in the `after_routes` method of trait `Hooks`.  Also implemented pub use of axum_csrf structs in module `auth`.


kaplanelad (commented):
Thanks @floscodes,

I have a few questions:

1. Why do you think this should be part of Loco’s default middleware?
2. Is this also relevant for REST APIs?
3. What if a user doesn’t want to use the default middleware and prefers to customize it, as the crate allows?
4. Wouldn't it be better if users integrated it directly into their projects, like you did, rather than having it included in the framework by default?

floscodes:
Thanks for the comment, @kaplanelad 

Here are my thoughts regarding default-CSRF-protection:

Since other frameworks that promote rapid development, such as Django, implement CSRF protection by default, I thought this would be a useful feature. Furthermore, it would live up to Loco's claim as a "one-person framework" in that you don't have to manually implement this feature, which is important for many applications, and development is faster. Regarding your question about an API: It's true that this isn't absolutely necessary in such use cases, but the advantage of the axum_csrf crate is that the CSRF token is only checked if you pass it as an argument to the respective handler function. If you don't do that, an API without CSRF protection will also work. If you want to customize CSRF protection despite the default implementation, you can simply override the axum `Router` in the `after_routes` method of the `Hooks` trait, thus having customized CSRF protection in your middleware.

PR #1489, opened 2025-06-20
Pin SeaORM version to "~1.1"

billy1624:
Hello Loco team, we're about to release SeaORM 1.2. It's a good idea to pin SeaORM dependency to "~1.1" to avoid auto upgrade.


kaplanelad:
@billy1624 Please correct me if I'm wrong, but if I specify 1.1.0 in Cargo.toml, only patch versions will be allowed (e.g., 1.1.x), and it will never automatically upgrade to 1.2.x. So pinning to 1.1.0 effectively prevents bumping to 1.2.x.

billy1624:
No, 1.1.0 will auto upgrade to 1.x

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax

```
1.2.3  :=  >=1.2.3, <2.0.0
1.2    :=  >=1.2.0, <2.0.0
1      :=  >=1.0.0, <2.0.0
0.2.3  :=  >=0.2.3, <0.3.0
0.2    :=  >=0.2.0, <0.3.0
0.0.3  :=  >=0.0.3, <0.0.4
0.0    :=  >=0.0.0, <0.1.0
0      :=  >=0.0.0, <1.0.0
```

~1.1 will do the trick

```
~1.2.3  := >=1.2.3, <1.3.0
~1.2    := >=1.2.0, <1.3.0
~1      := >=1.0.0, <2.0.0
```

DenuxPlays:
Quick question.
A minor version should introduce breaking changes why do we need to pin it?

PR #1501, opened 2025-07-04
feat: provide user creation via CLI

floscodes:
# Feature: Add users via CLI

By adding this feature, you can add a user in your current loco project via CLI.

Just change to the directory of your loco project and execute the following command:

`cargo loco createuser`

Then a dialogue will guide you.

```
👤 ❯ Enter username
📧 ❯ Enter email
🔒 ❯ Enter password
⚠️ ❯ Confirm password
```

If everything works, the terminal will display

`✅ User created successfully!`


kaplanelad:
Thanks, @floscodes!
I appreciate the suggestion, but I don't think it will fit all use cases. If the user changes the model, such as adding required fields, renaming, or removing fields, this code might break or become invalid.

floscodes:
I know what you mean, @kaplanelad. Perhaps this functionality could be provided by creating a trait called `CliCreator` for example. This trait could be implemented by default with a derive macro for the `Model` in `models::_entities::users` in the base template. Some fields, such as `id` or `api_key`, could be ignored by the CLI creation dialog by adding a macro attribute called `#[no-cli]`, for example. Regardless of whether the user adds or removes fields from the `Model` struct, all fields would be considered by the CLI dialog unless the user specifies `#[no-cli]` for the selected field.

What do you think about this?

kaplanelad:
This sounds a bit complex to me.

Maybe we can introduce a concept called **"Popular Tasks"**, a curated list of common actions that users can easily import into their projects (via download, CLI, or manually). Once imported, the task is automatically added and ready to use.

Some examples could include:

1. Create a new user
2. Delete a user
3. List users who haven't verified their account
4. Handle new user onboarding

We could easily expand this list with other useful tasks that many users would want to automate.

What do you think?

floscodes:
This is a good idea. But how should this "popular tasks" concept be introduced? Via a command, e.g. `cargo loco add` or in the config file? E.g. 

```
tasks:
    cli_create_user
```

or via a feature that one can add as a dependency in Cargo.toml?

kaplanelad:
I don’t think the base code of the tasks should be part of the framework itself.
Instead, we could manage the task code in a separate repository, and have the framework handle downloading and integrating the task into the user’s project automatically.

floscodes:
I see. So one repo called e.g. "loco-task" or "loco-feature" that can be compiled to a binary that will install all future tasks or features?

kaplanelad:
I don't think this needs to be a dedicated executable binary. I see it more like the existing injector we have, but instead of embedding the content directly in the framework at compile time, it should fetch it remotely at runtime.

floscodes:
I see. So I will think about how to build this...

kaplanelad:
Cool, closing this PR for now. Feel free to open an issue if you'd like to discuss implementation details or design ideas. I'd be happy to continue the conversation there. Thanks!

floscodes:
Ok! 👍

PR #1502, opened 2025-07-04
feat: add csrf protection by default

floscodes:
# Add CSRF protection by default

**Reopen of PR #1484 with separate branch**

I thought that this would be a useful feature.

Since other frameworks that promote rapid development, such as Django, implement CSRF protection by default, I thought this would be a useful feature. Furthermore, it would live up to Loco's claim as a "one-person framework" in that you don't have to manually implement this feature, which is important for many applications, and development is faster. Regarding your question about an API: It's true that this isn't absolutely necessary in such use cases, but the advantage of the axum_csrf crate is that the CSRF token is only checked if you pass it as an argument to the respective handler function. If you don't do that, an API without CSRF protection will also work. If you want to customize CSRF protection despite the default implementation, you can simply override the axum Router in the after_routes method of the Hooks trait, thus having customized CSRF protection in your middleware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ license = "Apache-2.0"
 
 [package]
 name = "loco-rs"
-version = "0.15.0"
+version = "0.16.0"
 description = "The one-person framework for Rust"
 homepage = "https://loco.rs/"
 documentation = "https://docs.rs/loco-rs"
@@ -52,7 +52,7 @@ integration_test = []
 embedded_assets = []
 
 [dependencies]
-loco-gen = { version = "0.15.0", path = "./loco-gen" }
+loco-gen = { version = "0.16.0", path = "./loco-gen" }
 backtrace_printer = { version = "1.3.0" }
 
 # cli
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,6 +121,7 @@ axum-test = { version = "17.0.1", optional = true }
 tree-fs = { version = "0.3", optional = true }
 
 chrono = { workspace = true }
+time = "0.3.41"
 
 uuid = { version = "1.10.0", features = ["v4", "fast-rng"] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ async-trait = { workspace = true }
 
 axum = { workspace = true }
 axum-extra = { version = "0.10", features = ["cookie"] }
+axum_csrf = { version = "0.11.0", features = ["layer"] }
 regex = { workspace = true }
 # mailer
 tera = { workspace = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,6 +156,7 @@ redis = { version = "0.31", features = ["aio", "tokio-comp"], optional = true }
 scraper = { version = "0.21.0", features = ["deterministic"], optional = true }
 
 dashmap = "6"
+base64 = "0.22.1"
 
 [workspace.dependencies]
 tera = { version = "1.19.1" }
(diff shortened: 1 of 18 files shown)


kaplanelad:
Let's continue the discussion in #1484. I think it would be better to include the loco new generate prject only for SaaS projects and exclude it from the API setup.

When it's part of the default Loco framework, it should be configurable via the config file (e.g. ability to disable it or customize settings), and it should also maintain versioning.

If we add it to the template, the user will receive it by default and take responsibility for managing it as needed.

floscodes:
Dear @kaplanelad,

I see – so I would say the ability to activate it in the config file is even better, because you don’t have to decide whether you want it or not when creating a new project…

floscodes:
@kaplanelad 
I refactored the code in the PR so that CSRF protection now in the config yaml file.

floscodes:
I still plan to add a few additional features so that you can, for example, customize the properties of the CSRF cookie.

floscodes:
# Customized CSRF config is now provided

## in the config yaml file (e.g. development.yaml):

### Example:

```
csrf_protection:
    enable: true

    cookie:
        name: "csrf-cookie"                         # Name of the CSRF cookie
        domain: "example.com"                  # Domain for the CSRF cookie, leave empty for default
        path: "/"                                             # Path for the CSRF cookie, leave empty for default
        same_site: Lax                                 # SameSite attribute for the CSRF cookie, options: Lax, Strict, None
        http_only: true                                 # Whether the CSRF cookie should be HTTP only
        lifetime: 3600                                  #Lifetime in seconds
        secure: true                                     # Whether the CSRF cookie should be secure (only sent over HTTPS)

     token:                                                  # CSRF token configuration    
        salt: "7f3d2b1e9c8a4f6d"            # Salt for CSRF token generation, should be a random string
        prefix_with_host: true                    # Whether to prefix the CSRF token with the host
        token_length: 32.                           # Length of the CSRF token

```

You can also skip one of these options; in this case, the default value of `CsrfConfig` from the crate `axum_csrf` will be used.

So you can use the default configuration by just writing

```
csrf_protection:
    enable: true
```

## Protect an endpoint

If you want to set the CSRF Token and render it into a template you can get a `String`of the token by do the following:

 - Add the Token to your controller
    ```rust
    use loco_rs__prelude::*;
    use csrf::CsrfToken;

    pub async fn index(token: CsrfToken) -> Result<Response>
    ```

 - Render a String of the CSRF Token into your template
    ```rust
    let token_string = token.authenticity_token().unwrap()
    ```

If you want to protect an endpoint do the following:

 - Pass the token to the endpoint-controller:
  ```rust
   pub async fn protected(token: CsrfToken) -> Result<Response>
  ```

 - Verify the token sent by the client by doing this:
  ```rust
   token.verify(<TOKEN SENT BY CLIENT>)
  ```

Cargo.toml
```diff
@@ -120,6 +121,7 @@ axum-test = { version = "17.0.1", optional = true }
 tree-fs = { version = "0.3", optional = true }
 
 chrono = { workspace = true }
+time = "0.3.41"
```
kaplanelad, line 124: Why not use chrono?

floscodes: The `axum_csrf` crate uses another [`Duration`](https://docs.rs/time/latest/src/time/duration.rs.html#43-50) struct that distincts from the one in the `chrono` crate. I think this is because the `Duration` struct in the `time` crate also allows [negative durations](https://docs.rs/time/latest/time/struct.Duration.html).

This is why I had to use this crate…

kaplanelad: Hey @floscodes,

I still feel that adding this crate, especially alongside another dependency isn't worth the tradeoff.
Implementing it manually in the `after_route` (as I initially suggested) would provide more value to users, since it gives them full control over configuration and crate versions. We could even publish a short guide on how to do it.

Given that, I think it's better to close this PR.

floscodes: Ok. So maybe there could be a short guide on the docs side...

PR #1512, opened 2025-07-12
Remove Migrator from boot_test_* doc comments

ellnix:
I've been learning loco recently and I noticed this small problem that the doc comments do not match the real function signature on these functions.

Hopefully, in the future you can expect contributions of value from me as I learn and use this excellent framework :laughing:

### DIFF
--- a/src/testing/request.rs
+++ b/src/testing/request.rs
@@ -173,11 +173,10 @@ pub async fn get_available_port() -> i32 {
 /// ```rust,ignore
 /// use myapp::app::App;
 /// use loco_rs::testing::prelude::*;
-/// use migration::Migrator;
 ///
 /// #[tokio::test]
 /// async fn test_create_user() {
-///     let boot = boot_test::<App, Migrator>().await;
+///     let boot = boot_test::<App>().await;
 /// }
 /// ```
 ///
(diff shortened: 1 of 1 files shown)


kaplanelad:
This change will be relevant only if you choose without the DB. If Loco runs with db the migration is relevant

ellnix:
@kaplanelad That is true for `boot` but not for `boot_test` as you can see from auto generated test files:

<details>
<summary>A brand new loco app's user test file</summary>

```text
$ loco new
✔ ❯ App name? · myapp
✔ ❯ What would you like to build? · Rest API (with DB and user auth)
✔ ❯ Select a DB Provider · Postgres
✔ ❯ Select your background worker type · Async (in-process tokio async tasks)

🚂 Loco app generated successfully in:
/tmp/myapp

- database: You've selected `postgres` as your DB provider (you should have a postgres instance to connect to)

$ head myapp/tests/models/users.rs -n 30
use chrono::{offset::Local, Duration};
use insta::assert_debug_snapshot;
use loco_rs::testing::prelude::*;
use myapp::{
    app::App,
    models::users::{self, Model, RegisterParams},
};
use sea_orm::{ActiveModelTrait, ActiveValue, IntoActiveModel};
use serial_test::serial;

macro_rules! configure_insta {
    ($($expr:expr),*) => {
        let mut settings = insta::Settings::clone_current();
        settings.set_prepend_module_to_snapshot(false);
        settings.set_snapshot_suffix("users");
        let _guard = settings.bind_to_scope();
    };
}

#[tokio::test]
#[serial]
async fn test_can_validate_model() {
    configure_insta!();

    let boot = boot_test::<App>() // <--- no migrator
        .await
        .expect("Failed to boot test application");

    let invalid_user = users::ActiveModel {
        name: ActiveValue::set("1".to_string()),
```

</details>

`boot_test` simply calls your `App`'s `boot` where `Migrator` is already injected. Moreover you can just look at the function signature directly under the comment:

```rust
pub async fn boot_test<H: Hooks>() -> Result<BootResult>
```

PR #1538, opened 2025-07-20
Fixes to syntax highlighting on docs site

ethanwood17:
Summary
- added a custom Terraform syntax highlighting file (terraform and hcl syntax highlighting is not included in Zola by default)
- fixed references to some syntaxes (shell -> sh)

### DIFF
--- a/docs-site/config.toml
+++ b/docs-site/config.toml
@@ -33,6 +33,7 @@ highlight_themes_css = [
   { theme = "OneHalfDark", filename = "syntax-theme-dark.css" },
   { theme = "OneHalfLight", filename = "syntax-theme-light.css" },
 ]
+extra_syntaxes_and_themes = ["syntaxes"]
 [extra]
 # Put all your custom variables here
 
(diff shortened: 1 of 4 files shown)


docs-site/config.toml
```diff
@@ -33,6 +33,7 @@ highlight_themes_css = [
   { theme = "OneHalfDark", filename = "syntax-theme-dark.css" },
   { theme = "OneHalfLight", filename = "syntax-theme-light.css" },
 ]
+extra_syntaxes_and_themes = ["syntaxes"]
```
kaplanelad, line 36: can you show before and after this change?

ethanwood17: Before
Screenshot 2025-07-20 at 15 37 36

After
Screenshot 2025-07-20 at 15 38 23

Adding custom syntaxes to highlight is documented [here](https://www.getzola.org/documentation/content/syntax-highlighting/)

kaplanelad: ok and why we need `docs-site/syntaxes/Terraform.sublime-syntax` file? it not should be in zola?

PR #1546, opened 2025-07-24
feat: decouple JWT authentication from database dependency

BigtoC:


kaplanelad (commented):
Hey @BigtoC,
Could you also take care of the feature dependencies in [prelude.rs](https://github.com/loco-rs/loco/blob/master/src/prelude.rs#L22C22-L22C30)?

BigtoC:
> Hey @BigtoC, Could you also take care of the feature dependencies in [prelude.rs](https://github.com/loco-rs/loco/blob/master/src/prelude.rs#L22C22-L22C30)?

Thank you for pointing this out!!

Issue #1550, opened 2025-07-25
[Bug] build --release fails with version 0.16


BigtoC:
**Description**

I use the CLI (version **0.16.1**) to generate a minimal service
```shell
loco new
✔ ❯ App name? · test-template
✔ ❯ What would you like to build? · lightweight-service (minimal, only controllers and views)

🚂 Loco app generated successfully in:
```

When I run `cargo build --release`, I get an error and fail to build:
```shell
error[E0063]: missing field `tera_post_process` in initializer of `TeraView`
   --> /Users/bigto/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/loco-rs-0.16.2/src/controller/views/engine.rs:113:12
    |
113 |         Ok(Self {
    |            ^^^^ missing `tera_post_process`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `loco-rs` (lib) due to 1 previous error
```

<img width="1848" height="1186" alt="Image" src="https://github.com/user-attachments/assets/c73d72bb-6ac4-488c-949f-13de7561eafc" />

But I can build if I change the version to **0.15**

<img width="1814" height="1954" alt="Image" src="https://github.com/user-attachments/assets/2e16463f-ce75-4518-a4ff-36c3d15d0d2a" />

**To Reproduce**

1. Run loco new to generate a minimal service
2. Make sure `loco-rs` version is "0.16"
3. Run `cargo build --release`

**Expected Behavior**

Build successfully with loco version = "0.16"

**Environment:**
loco cli: 0.16.1

```toml
[workspace.dependencies]
loco-rs = { version = "0.16", default-features = false }
```

```shell
rustc --version
rustc 1.88.0 (6b00bc388 2025-06-23)
```

**Additional Context**

schungx:
Oops. I might have missed something in the release build.

BigtoC:
> Oops. I might have missed something in the release build.

@schungx Thanks for your hints! I opened a PR for this: https://github.com/loco-rs/loco/pull/1551. Could you take a look?

schungx:
I don't think you can remove that. Post processing is required even for release builds in order to handle i18n.

I suggest removing line 116 instead.

Pokymon:
I got the same issue.

kaplanelad:
Hotfix released in version 0.16.3.

PR #1551, opened 2025-07-25
fix: add missing debug_assertions for tera_post_process

BigtoC:
Fix bug: https://github.com/loco-rs/loco/issues/1550

### DIFF
--- a/src/controller/views/engine.rs
+++ b/src/controller/views/engine.rs
@@ -19,7 +19,6 @@ pub struct TeraView {
     #[cfg(debug_assertions)]
     pub view_dir: String,
 
-    #[cfg(debug_assertions)]
     pub tera_post_process: Option<TeraPostProcessor>,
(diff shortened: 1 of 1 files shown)


schungx:
Post process cannot be removed. Otherwise the template-driven code won't compile since it is needed for i18n. 

Remove the attribute on line 116 instead.

BigtoC:
> Post process cannot be removed. Otherwise the template-driven code won't compile since it is needed for i18n.
> 
> Remove the attribute on line 116 instead.

Do you mean remove this line?
```rust
let ctx = tera::Context::default();
```
image

And rollback all `#[cfg(debug_assertions)]` I added for `tera_post_process`?

schungx:
This looks good. Suggest this to be merged and a new patch release made to correct my stupid mistake in `0.16.2`.

PR #1578, opened 2025-08-06
docs: enhance JWT configuration examples in config.rs

BigtoC:


kaplanelad:
Hey @BigtoC, I’d prefer to keep the `example curl request with JWT token` in the documentation website rather than in the library itself. The URL depends on the user and may change, which could make the library docs inaccurate.

PR #1582, opened 2025-08-08
Add pre-commit hooks for style check

BigtoC:
Run style check before commit to find out any style check issue before ci fails.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,15 @@
+echo "Running cargo fmt check..."
+cargo fmt --all -- --check
+if [ $? -ne 0 ]; then
+    echo "cargo fmt failed"
+    exit 1
+fi
+
+echo "Running cargo clippy..."
+cargo clippy --all-features -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms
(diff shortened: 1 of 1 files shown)


BigtoC:
weird, `cargo clippy --all-features -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms` didn't fail in local.

kaplanelad:
It is depends of your local rust version.

NexVeridian:
I don't really like pre-commit hooks, but the correct command is here, and requires nightly for some of the formats:

also don't put `rustup default nightly`or `+nightly` in this pr since that'll fail if installing nightly from not rustup ( nix fenix / oxalica/rust-overlay )

https://github.com/loco-rs/loco/blob/02fec6cd89f89d5a0e7e471bd7a63bdb75b42e49/CONTRIBUTING.md?plain=1#L48-L56

kaplanelad:
+1. 
I prefer not to include pre-commit hooks in the project itself.
Developers can choose to set them up locally if they want, but it shouldn't be enforced at the repo level.

BigtoC:
Thank you for your inputs! I'll close this PR!

PR #1599, opened 2025-08-23
Feature/enable edit cors expose headers

ofabianomartins:
I was developing an app with some custom response headers. These headers were used to provide the total count of requests in a table and when returning the rows of a paginated search. 

After returning the data, I noticed that the CORS middleware did not return the `Access-Control-Expose-Headers`, and it was not possible to configure this in the config options. This PR introduces a configuration option `expose_headers` to allow setting custom response headers in the CORS middleware.

```
server:
  ...
  middlewares:
    ...
    cors:
      enable: true
      expose_headers:  # -> this is the new option
         - X-CUSTOM-HEADER1
         - X-CUSTOM-HEADER2
```

### DIFF
--- a/docs-site/content/docs/the-app/controller.md
+++ b/docs-site/content/docs/the-app/controller.md
@@ -335,7 +335,7 @@ This is the stack in `development` mode:
 $ cargo loco middleware --config
 
 limit_payload          {"body_limit":{"Limit":1000000}}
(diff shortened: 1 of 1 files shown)


kaplanelad (commented):
Thanks, @ofabianomartins!
Could you also add the documentation here:
https://github.com/loco-rs/loco/blob/d361e43fb1fbe29ad1638d7ffb3150df3eb04e34/docs-site/content/docs/the-app/controller.md?plain=1#L852

ofabianomartins:
I added this documentation in my PR. You referred to merged commit. I don’t understand what you are asking for.

PR #1600, opened 2025-08-23
Feature/adding new imports about multipart

ofabianomartins:
I was developing an app to upload files and noticed that `Multipart` and `Query` is not defined in `use loco_rs::prelude::*`. So a added to `src/prelude.rs` this two `use axum::extract::Multipart` and  `use axum::extract::Query`. 

```
#![allow(clippy::missing_errors_doc)]
#![allow(clippy::unnecessary_struct_initialization)]
#![allow(clippy::unused_async)]
use axum::debug_handler;
use loco_rs::prelude::*;

use crate::models::_entities::users;
use crate::services::StorageService;

#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct ResponseUpload {
    pub uuid: String,
    pub status: String
}

impl ResponseUpload {
    pub fn success(uuid: String) -> Self {
        Self { status: "success".to_string(), uuid: uuid }
    }
}

#[debug_handler]
pub async fn upload(
    _auth: auth::JWTWithUser<users::Model>,
    State(ctx): State<AppContext>,
    mut multipart: Multipart
) -> Result<Response> {
    let storage_service = StorageService::build(&ctx);
    while let Some(field) = multipart
        .next_field()
        .await
        .map_err(|e| Error::BadRequest(e.to_string()))?
    {
        if let Some(file_name) = field.file_name().map(|f| f.to_string()) {
            let data: Vec<u8> = field
                .bytes()
                .await
                .map_err(|e| Error::BadRequest(e.to_string()))?
                .to_vec();

            let doc = storage_service.save_dir(file_name, &data).await.unwrap();

            return format::json(ResponseUpload::success(doc.uuid));
        }
    }

    Err(Error::BadRequest("No file uploaded".into()))
}

pub fn routes() -> Routes {
    Routes::new()
        .prefix("api/documents/")
        .add("upload", post(upload))
}
```

```

```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ thiserror = "1"
 serde = "1"
 serde_json = "1"
 async-trait = { version = "0.1.74" }
-axum = { version = "0.8.1", features = ["macros"] }
+axum = { version = "0.8.1", features = ["macros", "multipart"] }
 tower = "0.4"
 tower-http = { version = "0.6.1", features = [
     "trace",
--- a/docs-site/content/docs/extras/pluggability.md
+++ b/docs-site/content/docs/extras/pluggability.md
@@ -885,7 +885,7 @@ You have two main ways to retrieve data in your controllers:
   use loco_rs::prelude::*;
   use crate::app::MyClonableService; // Or wherever it's defined
 
-  #[axum::debug_handler]
+  #[debug_handler]
   pub async fn index(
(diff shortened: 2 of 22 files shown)


kaplanelad (approved):

kaplanelad:
Can you please fix the CI?

ofabianomartins:
I updated the tests and ran them on my local machine. They work!

PR #1607, opened 2025-08-30
Feature/adding new templates to migrations

ofabianomartins:
There is no bug detected and fixed. Just adding a new template when adding a index and creating a function to wrap the add index function.

``` 
///
/// Add a index to a table with a list of column
///
/// ```ignore
/// add_column(m, "movies", "title", Vec<String>).await;
/// ```
/// # Errors
/// fails when it fails
pub async fn add_index(
    m: &SchemaManager<'_>,
    table: &str,
    name: &str,
    column_names: &[&str],
) -> Result<(), DbErr> {
    let nz_table = normalize_table(table);
    let mut index = Index::create();
    index.name(name.to_string()).table(Alias::new(nz_table));

    for column_name in column_names {
        index.col(Alias::new(*column_name));
    }

    m.create_index(index.if_not_exists().to_owned()).await?;
    Ok(())
}

///
/// Drop a index from a table.
///
/// ```ignore
/// drop_column(m, "movies", "title").await;
/// ```
/// # Errors
/// fails when it fails
pub async fn remove_index(m: &SchemaManager<'_>, table: &str, name: &str) -> Result<(), DbErr> {
    let nz_table = normalize_table(table);
    m.drop_index(
        Index::drop()
            .name(name.to_string())
            .table(Alias::new(nz_table))
            .if_exists()
            .to_owned(),
    )
    .await?;
    Ok(())
}
```

I added this two functions on src/schema.rs and create two functions on loco-gen.

### DIFF
--- a/docs-site/content/docs/the-app/models.md
+++ b/docs-site/content/docs/the-app/models.md
@@ -37,7 +37,6 @@ Your local postgres database should be with <code>loco:loco</code> and a db name
 For your convenience, here is a docker command to start up a Postgresql database server:
 
 <!-- <snip id="postgres-run-docker-command" inject_from="yaml" template="sh"> -->
-
 ```sh
 docker run -d -p 5432:5432 \
   -e POSTGRES_USER=loco \
@@ -45,7 +44,6 @@ docker run -d -p 5432:5432 \
   -e POSTGRES_PASSWORD="loco" \
   postgres:15.3-alpine
 ```
-
 <!-- </snip> -->
 
 Finally you can also use the doctor command to validate your connection:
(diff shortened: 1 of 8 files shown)


ofabianomartins:
When I ran cargo clippy --fix, it fixed a lot of files. After that, a problem occurred when installing `sea-orm-cli` in the pipeline.

kaplanelad:
I noticed that this PR includes many file changes that don’t seem related to its purpose.  
Could you update it to include only the relevant files?

ofabianomartins:
Of course, but when I do this, fmt shows problems in other files. So I run Clippy to auto-fix the issues. I will revert to the essential changes, but you will see what happens.

ofabianomartins:
I understood what happened! The last modification added debug_handler to loco::prelude, but the wizard test is executed using version which doesn't include this modification. I added use axum::debug_handler to make the test pass, but a better solution is needed.

kaplanelad:
Thanks @ofabianomartins for this PR!
Could you please make sure it only includes the relevant changes?

ofabianomartins:
I will refactor the PR. I believe that was other things to do. I have a ideia to implement all method from rails active record. I believe substitute this function by macros is a good ideia.

kaplanelad:
Hey @ofabianomartins,
I’d appreciate it if you could also add this to the docs here:
https://github.com/loco-rs/loco/blob/master/docs-site/content/docs/the-app/models.md#migrations

ofabianomartins:
I will refactor all PR to use macros.

PR #1614, opened 2025-09-06
docs: creating a the custom extractor in documentation

ofabianomartins:
Just adding a specific documentation to a problem i had and spend a lot of time. I tried to set in most correct place in documentation.


ofabianomartins:
@kaplanelad , there is any problem here? It's just a comment.

PR #1632, opened 2025-09-21
Update rhai

kaplanelad:


schungx:
```rust
if settings.asset != () && settings.asset.is_server_side {
```

can be more compactly expressed as:

```rust
if settings.asset?.is_server_side ?? false {
```

Etc.

PR #1634, opened 2025-09-25
Upgrade SeaORM to 2.0

tyt2y3:
Hi! As you may know SeaORM 2.0 is now in rc status. There is not much breaking changes that affected Loco, and we've been using this fork to test things, and it's also tested as part of our CI process.

Nothing to be done about it right now, other than being a reference.


kaplanelad:
Hey @tyt2y3, I noticed 2.x is still an RC version, let’s wait for the stable release before upgrading.

PR #1642, opened 2025-09-28
feat: creating UuidUniqWithDefault and UuidWithDefault

ofabianomartins:
I need to add a default value to uuid column and add a default value to uniq uuid column.


src/storage/stream.rs
```diff
@@ -61,6 +61,7 @@ impl Stream for BytesStream {
 impl BytesStream {
     /// Convert the `BytesStream` into an axum `Body` for HTTP responses.
     /// This enables zero-copy streaming directly to the HTTP response.
+    #[allow(clippy::double_must_use)]
```
kaplanelad: Please remove it as well

ofabianomartins:
Changes applied!

loco-new/base_template/src/controllers/home.rs
```diff
@@ -1,4 +1,5 @@
 use loco_rs::prelude::*;
+use axum::debug_handler;
```
kaplanelad: I don’t think this is related.
Since `axum::debug_handler` is now part of `loco_rs::prelude::*`; you can safely remove the import.

kaplanelad (approved):

ofabianomartins:
Hi @kaplanelad, you approved but not merge the PR. Something wrong?

Issue #1650, opened 2025-10-06
cannot find attribute debug_handler in this scope


Sirazulhaq-Abdul-Majid:
**Description**

Following the guide on the https://loco.rs/docs/the-app/your-project/ page and tried to cargo loco generate model posts. And failed because of debug_handler

```
error: cannot find attribute debug_handler in this scope
  --> src/controllers/auth.rs:45:3
   |
45 | #[debug_handler]
   |   ^^^^^^^^^^^^^
   |
help: consider importing this attribute macro
   |
1  + use axum::debug_handler;
   |

error: cannot find attribute `debug_handler` in this scope
  --> src/controllers/auth.rs:76:3
   |
76 | #[debug_handler]
   |   ^^^^^^^^^^^^^
   |
help: consider importing this attribute macro
   |
1  + use axum::debug_handler;
   |

error: cannot find attribute `debug_handler` in this scope
  --> src/controllers/auth.rs:97:3
   |
97 | #[debug_handler]
   |   ^^^^^^^^^^^^^
   |
help: consider importing this attribute macro
   |
1  + use axum::debug_handler;
   |

error: cannot find attribute `debug_handler` in this scope
   --> src/controllers/auth.rs:119:3
    |
119 | #[debug_handler]
    |   ^^^^^^^^^^^^^
    |
help: consider importing this attribute macro
    |
1   + use axum::debug_handler;
    |

error: cannot find attribute `debug_handler` in this scope
   --> src/controllers/auth.rs:136:3
    |
136 | #[debug_handler]
    |   ^^^^^^^^^^^^^
    |
help: consider importing this attribute macro
    |
1   + use axum::debug_handler;
    |

error: cannot find attribute `debug_handler` in this scope
   --> src/controllers/auth.rs:161:3
    |
161 | #[debug_handler]
    |   ^^^^^^^^^^^^^
    |
help: consider importing this attribute macro
    |
1   + use axum::debug_handler;
    |

error: cannot find attribute `debug_handler` in this scope
   --> src/controllers/auth.rs:229:3
    |
229 | #[debug_handler]
    |   ^^^^^^^^^^^^^
    |
help: consider importing this attribute macro
    |
1   + use axum::debug_handler;
    |

error: could not compile `todolist` (lib) due to 7 previous errors
warning: build failed, waiting for other jobs to finish...
```

**To Reproduce**

[...]

PR #1654, opened 2025-10-08
fix: missing axum import in auth template

ltabis:
After trying out the base template of loco and running `cargo loco start`, I got the same error as found in the issue #1650. I believe this is a simple missing import. i don't have time to dig deeper in the tests right now, but are tests that install the base loco project template missing ? If so, would it be an interesting thing to add to the project ?

I tested my changes by running the app after fixing the import, works fine.

Btw, thank you for building this framework, it looks awesome from what I have seen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+
+## v0.16.4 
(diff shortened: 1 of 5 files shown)


blackwood:
Experienced the same issue, confirmed this fixed it.

kaplanelad:
Hey,  
we’ve added this import `use loco_rs::prelude::*;`, and it should be included in the latest version.  
Can you try upgrading and let me know if the issue still occurs?

ltabis:
@kaplanelad looking good, thank you ! Feel free to close this.

PR #1657, opened 2025-10-10
refactor: replace trait fn bounds with impl AsyncFn

alwayys-afk:
Some of the tests had this pattern:
```
async fn request_internal<F, Fut>(callback: F, boot: &BootResult, test_server_config: RequestConfig)
where
    F: FnOnce(TestServer, AppContext) -> Fut,
    Fut: std::future::Future<Output = ()>,
```
switched to this:
```
async fn request_internal(
    callback: impl AsyncFnOnce(TestServer, AppContext),
    boot: &BootResult,
    test_server_config: RequestConfig)
```


dsgallups (approved):
looks good, though the PR description doesn't quite match what this does. 
Before:
```rust
request::<App, _, _>(|test_server, etc)| async { ... });
```
This PR:
```rust
request::<App>(|test_server, etc)| async { ... });
```

jondot:
Thanks @alwayys-afk — I like the direction (dropping the trait fn bounds for `impl AsyncFn`), but as-is it reintroduces an E0282 inference failure in the generated auth tests, so I am deferring it past 1.0 rather than rejecting it. I would like to return to it once that is sorted. Appreciate the refactor and the patience!

PR #1658, opened 2025-10-11
Creating new options to with deal errors

ofabianomartins:
Just adding new type to error handling src/error.rs

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -128,7 +128,7 @@ pub fn bad_request<T: Into<String>, U>(msg: T) -> Result<U> {
 /// Currently this function doesn't return any error. this is for feature
 /// functionality
 pub fn not_found<T>() -> Result<T> {
-    Err(Error::NotFound)
+    Err(Error::NotFound("".to_string()))
 }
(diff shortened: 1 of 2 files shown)


kaplanelad:
Can you provide more details about the problem and what you’re trying to achieve?

ofabianomartins:
It's not finished! I want to add messages to the error responses. For example, 'Bad Request' doesn't explain what is wrong with the call — it only shows 'Bad Request' in the body. That's not a good approach. When I develop an app, I always display the message from the response body on the screen. Just showing 'Bad Request' isn’t helpful.

kaplanelad:
I’m not comfortable returning internal errors to users due to security concerns.
If a custom error/message needs to be returned, it should be done via RenderBuilder in [format.rs](https://github.com/loco-rs/loco/blob/master/src/controller/format.rs), with full responsibility for the output on the implementer.

ofabianomartins:
I declining this PR. It's useless!

schungx:
> it should be done via RenderBuilder in format.rs,

First time I know of `RenderBuilder`!

PR #1659, opened 2025-10-11
override a secure header

maksimryndin:
Hey guys! thank you for the awesome framework!

There are cases when I want to change dynamically a CSP header (e.g. when inlining styles into the template).
To uphold CSP I need to render template with a nonce (or sha256) value.

The change to secure_headers middleware - when some secure header is already set by a controller, then the default one is not applied

### DIFF
--- a/src/mailer/email_sender.rs
+++ b/src/mailer/email_sender.rs
@@ -3,7 +3,7 @@
 //! sending emails with options like sender, recipient, subject, and content.
 
 use lettre::{
-    message::MultiPart,
+    message::{header, MultiPart},
     transport::smtp::{authentication::Credentials, extension::ClientId},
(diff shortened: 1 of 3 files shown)


kaplanelad:
thanks @maksimryndin. 

Can you check the test that you added?

maksimryndin:
> thanks @maksimryndin.
> 
> Can you check the test that you added?

Hey @kaplanelad ! Sorry for the long reply, I will check the test in some time (just overloaded with other things atm)

maksimryndin:
@kaplanelad hey :) The second clippy issue comes from the other (unrelevant to the PR) file (https://github.com/loco-rs/loco/blob/master/src/storage/stream.rs#L64). Merely adding `must_use` seems a false positive for me, wdyt?

kaplanelad (approved):

PR #1662, opened 2025-10-15
Allow handlers to take `Option<JWT>` as a parameter

pantsman0:
This PR adds an implementation of `OptionalFromRequestParts<S> for  JWT`, allowing handler functions in controllers such as `pub async fn render_home(jwt: Option<auth::JWT>, ViewEngine(v): ViewEngine<TeraView>) -> Result<Response>` for things like redirecting to the login page if the JWT is not present in a cookie.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -172,7 +172,7 @@ serde_json = "1"
 async-trait = { version = "0.1.74" }
 axum = { version = "0.8.1", features = ["macros", "multipart"] }
 tower = "0.4"
-tower-http = { version = "0.6.1", features = [
+tower-http = { version = "0.6.8", features = [
     "trace",
     "catch-panic",
     "timeout",
--- a/docs-site/content/docs/the-app/controller.md
+++ b/docs-site/content/docs/the-app/controller.md
@@ -1218,7 +1218,7 @@ impl From<notes::Model> for ListResponse {
 
 impl PaginationResponse {
     #[must_use]
-    pub fn response(data: PaginatedResponse<notes::Model>, pagination_query: &PaginationQuery) -> Pager<Vec<ListResponse>> {
+    pub fn response(data: PageResponse<notes::Model>, pagination_query: &PaginationQuery) -> Pager<Vec<ListResponse>> {
         Pager {
             results: data
                 .page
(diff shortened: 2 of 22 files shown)


askor:
Edit: see post below

This would be a big improvement for me. Not sure how else to create a frontpage on the html/htmx stack, with different state depending on the logged in status, apart from doing multiple calls.

askor:
I found a solution for this, as it's already solved in Axum by [handling extractor rejections with Result](https://docs.rs/axum/latest/axum/extract/index.html#handling-extractor-rejections).

So you can wrap JWT in Result like:

    pub async fn render_home(jwt: Result<auth::JWT>, ViewEngine(v): ViewEngine<TeraView>) -> Result<Response>
    
and handle the error if there is no auth present

pantsman0:
@askor yeah, I should have thought of that...

I think this is still OK though as it can bubble up JWT format/validity errors separate from JWT presence.

kaplanelad:
thanks @pantsman0,
I would love to see a test for it. can you add tests here: https://github.com/loco-rs/loco/tree/master/tests/controller/extractor

pantsman0:
@kaplanelad Can you provide a little more guidance? There are no existing JWT tests in that folder, are you OK for me to just have at it?

I have made changes to the tests in `src/controllers/extractor/auth.rs` as well, but these are just the JWTLocationConfig tests. Case 5 is currently failing, but this may need some more specific behaviour - case 5 searches multiple locations and used to give a message about the jwt configuration, but that's not actually something you want to send to users. Should we have some functionality there so it only shows the developer message when the app is running in development mode?

kaplanelad:
Sure, lt me open a pr that testing the current auth flow, then you can update your branch and adding yours

kaplanelad:
I’ve opened a PR for the auth tests: https://github.com/loco-rs/loco/pull/1671.
Once it passes CI, I’ll merge it into master, and then you can take it from there.

pantsman0:
@kaplanelad sorry for the ping, I don't know if you get one when I put an update.

I fixed the style CI fail.

kaplanelad:
The CI is still failing, so I’m closing this PR for now.
If it’s still relevant, please feel free to reopen it and fix the CI issues.
If you’re running into any problems, I’m happy to help.

PR #1672, opened 2025-10-22
Add automatic creation of open api documentation through the aide crate

Isaac-Leonard:
These changes allow for open api documentation to be generated automatically when people run the app.
I wrote these changes awhile ago hence being several versions behind and I don't fully remember how the changes work but I've been meaning to get around to making a PR for a while now.
I'm aware https://github.com/loco-rs/loco-openapi-Initializer has come about since I made this fork but that makes use of Utoipa which I have always found fiddly to use. It requires that paths and params be repeated twice, once in the documentation and once in the code, which defeats the entire point of auto generating documentation.
This PR instead uses the code as a single source of truth meaning that changes in the code are reflected in the documentation in real time.

I've furthermore been able to make use of this by using the generated docs to create a full api client in my frontend code using the openapi-qraft generator which may be worth integrating in a future PR.

I'm pretty certain this PR involves several breaking changes however as I said, it's been almost a year since I made the fork and I haven't worked on it since so I don't remember and won't have the time to check it properly for a couple of weeks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,7 @@ axum-test = { version = "17.0.1", optional = true }
 chrono = { workspace = true }
 cfg-if = "1"
 
-uuid = { version = "1.10.0", features = ["v4", "fast-rng"] }
+uuid = { version = "1.10.0", features = ["v4", "fast-rng", "serde"] }
 
 # File Upload
 opendal = { version = "0.50.2", default-features = false, features = [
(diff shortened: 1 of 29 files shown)


NexVeridian:
I don't think this pr is going to get merged, you can see the comments about the predecessor pr of https://github.com/loco-rs/loco-openapi-Initializer here #984

I think the best possibility of getting this added is to refactor it to a `Initializer`, see https://github.com/loco-rs/loco-openapi-Initializer, there is probably going to be some workarounds like a second parallel `ApiRouter`/`AxumRouter` and needing to add to both like [here](https://github.com/loco-rs/loco-openapi-Initializer?tab=readme-ov-file#automatically-adding-routes-to-the-openapi-spec-visualizer)

PR #1687, opened 2025-11-06
feat: optionally exclude port from server full_url

AnthonyMichaelTDM:
This pull request introduces a new configuration option to control whether the server port should be included in the generated server URL for mailers and such.
The default value for this field is true, which will mimic the previous behavior

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -445,6 +445,7 @@ pub enum JWTLocationConfig {
 /// server:
 ///   port: {{ get_env(name="NODE_PORT", default=5150) }}
 ///   host: http://localhost
+///   port_in_url: true
 ///   middlewares:
 ///     limit_payload:
 ///       enable: true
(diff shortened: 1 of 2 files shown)


AnthonyMichaelTDM:
I was going to open an issue for this but I figured a PR would be easier for everyone involved

kaplanelad:
Hey @AnthonyMichaelTDM, can you explain why do you need it?

AnthonyMichaelTDM:
To avoid including invalid URLs in auth emails in production.

AnthonyMichaelTDM:
Like, in dev I want https://localhost:8080, but in production https://example.com:8080 is incorrect in most cases

I've been using 

```rust
pub fn domain(ctx: &loco_rs::app::AppContext) -> String {
    match ctx.environment {
        loco_rs::environment::Environment::Production => ctx.config.server.host.clone(),
        _ => ctx.config.server.full_url(),
    }
}
```

As a workaround so this is more of a convenience thing than anything else

PR #1690, opened 2025-11-14
Build YAML by streaming

RandomInsano:
This PR changes `cargo loco seed --dump` to stream data to files instead of building a large string before writing to disk. It uses a buffered writer to lessen the access overhead. I will add the same for seeding, but that will require a bit more work and I wanted to get opinions on this first.

One problem is that I couldn't find a reasonable way to access sqlx's streaming for custom queries in SeaORM, so in my case Loco is still allocating 6GB of RAM for 3 million rows. Is it alright for me to access sqlx directly here?

Sorry for the formatting shift here. Let me know if you want me to revert it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -172,7 +172,7 @@ serde_json = "1"
 async-trait = { version = "0.1.74" }
 axum = { version = "0.8.1", features = ["macros", "multipart"] }
 tower = "0.4"
-tower-http = { version = "0.6.1", features = [
+tower-http = { version = "0.6.8", features = [
     "trace",
     "catch-panic",
     "timeout",
--- a/loco-gen/src/lib.rs
+++ b/loco-gen/src/lib.rs
@@ -90,7 +90,7 @@ impl Mappings {
             field,
             allow_fields
                 .iter()
-                .map(|&s| s.to_string())
+                .map(|&s| s.clone())
                 .collect::<Vec<String>>()
                 .join(",")
         ))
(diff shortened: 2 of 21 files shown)


RandomInsano:
Did some experimenting here. Consider this an RFC because I've fixed my problem and can polish this up to upstream.

Because serde_yaml didn't support streaming inside of async context with `serde_yaml::Serializer`, I changed the serializing to serde_json. The YAML serializer failed due to an internal `static lifetime, but it's also deprecated. Frankly, I prefer YAML for this context myself.

The format of the JSON is not a true JSON array but just a series of JSON objects side-by-side. I've confirmed `jq` is happy with this, and it allows us to use `serde_json::Deserializer::from_reader()` to read data in object by object. I'm not sure if it's safe/possible to put this into a database transaction, but happy to do that if the import should be atomic (say a dump file is corrupted).

I raised a `dump()` function into the `Hooks` trait so that the Generics could be used when serializing. What I don't like at the moment is that the implementor of that function needs to do the table filtering, but I don't have a good solution to promote it to `run_app_dump()`... Some advice there would be help, but I can mull it over if the work I've done here is worthwhile.

kaplanelad:
Hey @RandomInsano,
I noticed that this PR includes several changes that aren’t related to its main purpose.
Please make sure the CI checks pass before we proceed.

kaplanelad:
@RandomInsano, this PR contains many unrelated changes. I’m closing it for now.  
Please feel free to open a new PR that includes only the relevant changes.

PR #1693, opened 2025-11-18
Priority queue

jtwaleson:
Here we add a priority queue to the postgres bg worker. This way, we can prioritize jobs.

Now we can do `Worker::perform_later_with_priority(ctx, params, Some(100))` to schedule a high prio job or with `Some(0)` for a low prio job.

I brought this up in https://github.com/loco-rs/loco/discussions/1462

What I'm using this for: we sync and analyze 100s of git repos. Each one has 3 stages: fetch, quick analysis, thorough analysis. As we want to show the results of the quick analysis as soon as possible, we give the following prios: `fetch: 5`, `quick analysis: 10`, `thorough analysis: 1`. This way, we always fetch first, but after fetching, we do a quick analysis right away. When everything is fetched and "quick analyzed", we handle the `thorough analysis`.

Please let me know what you think, and if we would want this for all queue backends or if we are fine with just postgres. If we don't want this at all, also fine.


kaplanelad:
Yes, this feature is required across all backends.

jtwaleson:
> Yes, this feature is required across all backends.

Great, I'll see what I can do.

jtwaleson:
Hey @kaplanelad , I've added support for sqlite and redis, and rebased my commits. The redis implementation needed some work as the native LPOP operations are incompatible with a priority queue report, and the combination with tags made it still a bit more difficult, hence the updated Lua script.

While creating the test cases I found some flaky behavior but I've fixed all the issues.

I've been running in production with the postgres implementation for months now, but I've not battle-tested sqlt and redis.

I believe it's ready for review & merge.

kaplanelad:
The priority queue feature is valuable, but a few things need to be addressed:

 1. Redis breaking change: Migrating from Lists to Sorted Sets means existing jobs are lost on upgrade. Please add clear upgrade instructions to the CHANGELOG, users need to drain their queues before upgrading, or accept job
 loss.
 2. Priority ordering reliability: The scoring approach for Redis needs to guarantee correct ordering under all conditions (high priorities, equal priorities with different timestamps, negative priorities). Please add tests
 that verify ordering is deterministic across all three backends, and document the priority semantics (higher = more urgent? what's the valid range?).
 3. MailerWorker priority 100: Hardcoding email priority is opinionated. please make this configurable or at least document it clearly.

jtwaleson:
Alright, 1 & 3 are simple. For 2 I'll need a bit more time. Thanks for reviewing @kaplanelad .

By the way, my startup https://comper.io is now out of the shadows and I can proudly say that it is fully built on Loco, so I wanted to say thanks for creating & maintaining the project!

jtwaleson:
Fixed all three points @kaplanelad . Let me know what you think.

Btw noticed some failures while running `cargo test` in `loco-gen` and `loco-new` but these seem unrelated.

jondot:
Thanks @jtwaleson — 1.0.0 ships priority queues (ZSET-backed, with `perform_later_with_priority`), which covers the same ground, so I am closing this as superseded — it absolutely informed the direction. Appreciate both queue PRs!

Issue #1652, opened 2025-10-07
Support templating in emails?


jtwaleson:
Right now we use `Tera::one_off` in `/src/tera.rs`, and each render happens in isolation.

**Expected Behavior**

I'd like to add use a common email template and use `extends` for each individual email.

Are you open to supporting this? If so, I'll take a stab at an implementation.

jtwaleson:
Sorry if this was a bit too cryptic, it made sense in my head but I forgot to add a bit of context.

In tera you can use `extends` and `block` functionality to have re-usable components, but because we use `Tera::one_off` this system does not work, as Tera will have access to only small parts of the templates.

kaplanelad:
hey @jtwaleson
The email templates are supporting tera, see it [here](https://github.com/loco-rs/loco/blob/master/src/mailer/template.rs). 
And you can pass and data that you want from the code to the tera template

jtwaleson:
Sorry for the delay. Bit busy with the company.

I'd like to have a template email with the company logo and footer etc, and then much smaller emails with the subject + body for each of them.

Typically you do this with `{% block subject %}Default Subject{% endblock subject %}`. You store this in e.g. `mailers/base.t`.

Then in the child you can do `{% extends "base.t"%} {% block subject %}Overridden Subject {% endblock subject %}`

If you try this now in Loco, you get this: ```called `Result::unwrap()` on an `Err` value: Tera(Error { kind: MissingParent { current: "__tera_one_off", parent: "base.t" }, source: None })```

I'd be happy to contribute a patch to initialize a context in Tera, so that we can use the power of templating. Let me know if you're open to a PR.

jondot:
Yes — and it shipped in 1.0.0. Mailer templates now support Tera inheritance and shared layouts: `Mailer::mail_template_with_shared` / `Template::new_with_shared`, and `loco generate mailer` scaffolds a `src/mailers/shared/` base layout that individual emails `extends`. Exactly the common-template + `extends` flow you sketched. Thanks @jtwaleson for taking the stab at it (#1694). Closing as done.

PR #1694, opened 2025-11-19
Add ability to use Tera template inheritence in emails

jtwaleson:
This is a first draft for adding templates to emails.

I noticed that I'd like to use Tera's `extends` syntax in my emails so that I can have one file with the layout of the html email and can just use smaller ones with message-specific layout bits.

Here we add a `src/mailers/shared/{base.t,subject.t,text.t}` and add these to the Tera context, so that en email templates you can just use `{% extends "base.t" %}`.

Before this change, this is not possible as we use the `one_off` rendering method and Tera has very little context.

Please let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ async fn after_routes(&self, router: AxumRouter, _ctx: &AppContext) -> Result<Ax
 
 
 
-- Feat: Mailer templates now support full Tera template features including inheritance, blocks, and extends. Templates can also be shared across multiple mailers using `Template::new_with_shared()`. [https://github.com/loco-rs/loco/pull/TBD](https://github.com/loco-rs/loco/pull/TBD)
+- **BREAKING**: `Template::new()` signature changed from `Template::new(dir)` to `Template::new(dir)?` to support template inheritance validation. Mailer templates now support full Tera template features including inheritance, blocks, and extends. Templates can also be shared across multiple mailers using `Template::new_with_shared()`. [https://github.com/loco-rs/loco/pull/1694](https://github.com/loco-rs/loco/pull/1694)
 
 ## v0.16.4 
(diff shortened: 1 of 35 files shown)


jtwaleson:
This is related to #1652

kaplanelad:
Hey @jtwaleson, 
this pr ready for review?

jtwaleson:
> Hey @jtwaleson, this pr ready for review?

I think it's ready for your opinion: do the maintainers want this feature or not? ;)

If you do, then I'll get this in merge-able shape (it should be 90% there already) and check if it's fully backwards compatible.

kaplanelad (changes requested):
Overall, this looks good!
I’ve suggested a few small changes, mainly focused on simplifying the tests.

Also, what about upgrades? If I use Loco and update to a new version that includes email templates, will it work as expected, or are there breaking changes?

jtwaleson:
@kaplanelad moved to insta and added an upgrade guide. Everything should keep working as expected, just when you create new mailers, they will default to a structure that supports templates. Previous instances of mailers need to be updated manually.

kaplanelad:
The template inheritance feature is great and well-implemented. Since we're planning a breaking changes release, the Template::new() signature change is acceptable.

 Please:
 1. Add this to the CHANGELOG under breaking changes.
 2. Add an upgrade guide entry explaining how to update existing mailer code (Template::new(dir) → Template::new(dir)?).
 3. Document the new shared template feature with a short example in the upgrade guide.

jtwaleson:
@kaplanelad thanks, I addressed the changes. I force pushed because I noticed one commit added something and another removed it again, it was a bit messy.

jondot:
Thanks @jtwaleson — Tera template inheritance and shared layouts for mailers shipped in 1.0.0. Closing in favor of the 1.0.0 implementation, which this informed. Appreciate it!

PR #1702, opened 2025-12-11
Fix TeraView memory issue and update loco-new for new Rhai version

schungx:
This PR fixes the excessive memory usage of TeraView in release builds by `Arc`-wrapping the Tera instance.

In real-life testing with a Tera-heavy app reduced RAM footprint from 3GB+ to 20MB.

The newest version of Rhai has built-in support for `Option` types in `CustomType` derive, breaking code in `loco-new`.  This PR also fixes that.

Finally, some fixes are made to the code base to shut up clippy and make CI passes, which took many more commits than expected.

### DIFF
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -89,11 +89,8 @@ pub async fn start<H: Hooks>(
         let scheduler = scheduler::<H>(&boot.app_context, None, None, None)?;
         tokio::spawn(async move {
             let res = scheduler.run().await;
-            if res.is_err() {
-                error!(
-                    err = res.unwrap_err().to_string(),
-                    "error while running scheduler"
-                );
+            if let Err(err) = res {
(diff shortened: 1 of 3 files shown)


kaplanelad:
Hey @schungx,
Thanks for the PR! Could you please split it into two separate pull requests?

schungx:
> Thanks for the PR! Could you please split it into two separate pull requests?

Probably can do but I think I'm lost somewhere trying to shut up clippy... which seems to be a lost cause...

What do you suggest that I do?  How do I separate it?

schungx:
I'm not a Git wizard by any means...  so what do I do now?

This bloody clippy is so f***king annoying.

kaplanelad:
One or for the memory fix and other is the bump the Rhai.

one you doing it let me know, I’ll take a look and help to solve the clippy errors

schungx:
Let me try

PR #1707, opened 2025-12-15
Cargo format 20251215

RandomInsano:
I'm not sure if this is appreciated, but my IDE auto-formats as it goes based on Cargo's `fmt` functionality. Because this project has formatting expectations based on [.rustfmt.toml](https://github.com/loco-rs/loco/blob/master/.rustfmt.toml) this PR should bring code back into the expected format.

### DIFF
--- a/src/controller/middleware/secure_headers.rs
+++ b/src/controller/middleware/secure_headers.rs
@@ -224,7 +224,6 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::controller::format;
     use axum::{
         body::Body,
         http::{header::CONTENT_SECURITY_POLICY, HeaderMap, HeaderValue, Method, Response},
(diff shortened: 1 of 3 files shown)


schungx:
Shouldn't your IDE automatically use `.rustfmt.toml` when running fmt?

I have not seen any large scale re-formatting in my IDE (VScode)...

kaplanelad:
In https://github.com/loco-rs/loco/blob/master/.rustfmt.toml, we need to remove configuration options that are not supported in the stable version of rustfmt.

The goal is for the format file to include only settings available in the stable release.
@RandomInsano, can you please take care of this and revert all changes to the already formatted files?

RandomInsano:
Yeah, I can take that on. There are a lot of features I personally use from nightly, so I have forced VSCode to use it which is likely where this is happening. I can also add a warning to the build steps so the format stays consistent.

I'll set aside time later this week or in the new year.

PR #1708, opened 2025-12-16
feat/add-popular-tasks

floscodes:
This PR adds a user deletion functionality to popular tasks.


loco-new/base_template/src/tasks/user_delete.rs
```diff
@@ -0,0 +1,59 @@
+use loco_rs::prelude::*;
+use dialoguer::{theme::ColorfulTheme, Confirm};
```
kaplanelad: Do new Cargo projects include `dialoguer` by default?
If not, I’d prefer not to add it to the project (io::stdout).

kaplanelad (changes requested):

loco-new/base_template/src/tasks/user_create.rs
```diff
@@ -11,10 +11,11 @@ impl Task for UserCreate {
     fn task(&self) -> TaskInfo {
         TaskInfo {
             name: "user:create".to_string(),
-            detail: "Create a new user with email, name, and password. Sends welcome email and sets up email verification.\nUsage:\ncargo run task user:create email:user@example.com name:\"John Doe\" password:\"securepassword\"".to_string(),
+            detail: "Create a new user with email, name, and password. Sends welcome email and sets up email verification.\nUsage:\ncargo run task user:create".to_string(),
```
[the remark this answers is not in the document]
floscodes: by mistake, sorry. I'm going to add it again

loco-new/base_template/tests/tasks/user_delete.rs.t
```diff
@@ -0,0 +1,41 @@
...
+        }).await;
+
+    assert!(user.is_ok());
+
+    let vars = task::Vars::from_cli_args(vec![
+        ("email".to_string(), email.to_string()),
+    ]);
+
+    assert!(
+        run_task::<App>(&boot.app_context, Some(&"user:delete".to_string()), &vars)
```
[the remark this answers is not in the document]
floscodes: Exactly. I'm going to fix that.

floscodes:
Hi @kaplanelad,

I refactored the code and fixed the remaining issues.

loco-new/base_template/src/tasks/user_delete.rs
```diff
@@ -0,0 +1,69 @@
...
+        };
+
+        let user_to_delete = users::Model::find_by_pid(&app_context.db, &input).await?;
+
+        tracing::info!(
+            message = "User to delete",
+            user_email = &user_to_delete.email,
+            user_pid = user_to_delete.pid.to_string(),
+            "delete user via task"
+        );
```
kaplanelad: Instead of using `tracing::info!` here, I think we should just use `println!`.
If the log level is higher than `info`, the user would otherwise see the
confirmation prompt without any user details, which is not ideal for a
destructive action.

loco-new/base_template/src/tasks/user_delete.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        let mut confirm = String::new();
+
+        // If test ist running simulate confirmation
+        if let Ok(var) = env::var("TEST_CAN_RUN_USER_DELET_BY_PID") {
```
kaplanelad: don’t think we should rely on a test-only environment variable (TEST_CAN_RUN_USER_DELET_BY_PID) in production code. It couples test behavior into runtime logic, is easy to forget to unset, and can be risky for a destructive task.

A more idiomatic and safer approach would be to add an explicit CLI flag such as force:true to bypass the confirmation prompt. This is a common convention for destructive commands, keeps intent explicit, and allows tests to run without special env vars.

kaplanelad (changes requested):

floscodes:
Hi @kaplanelad,

is it ok like this?

loco-new/base_template/src/tasks/user_delete.rs
```diff
@@ -0,0 +1,61 @@
...
+            stdin.read_line(&mut confirm).map_err(|err| {
+                tracing::error!(
+                    message = err.to_string(),
+                    "could not read confirmation input"
+                );
+                Error::string(&format!("Failed to read confirmation input. err: {err}",))
+            })?;
+
+            if confirm.trim().to_lowercase() != "yes" {
+                println!("⛔ User deletion cancelled - nothing has been deleted!");
```
kaplanelad: remove the ⛔

kaplanelad (changes requested):

kaplanelad:
> Hi @kaplanelad,
> 
> is it ok like this?

looks better. thanks. 
can you check the ci?

floscodes:
Hi @kaplanelad,

just a quick heads-up: the CI is now passing and all tests are running successfully.

Thank you!

kaplanelad:
How group related to delete user task?

floscodes:
The `delete`-method from the `ActiveModelTrait` requires the groups table to exist:

```
let _deleted_user = user_to_delete
            .into_active_model()
            .delete(&app_context.db)
            .await
            .map_err(|err| {
                tracing::error!(message = err.to_string(), "could not delete user");
                Error::string(&format!("Failed to delete user. err: {err}",))
            })?;
```

otherwise the test fails with the following error:

```
Execution Error: error returned from database:
(code: 1) no such table: main.groups
```
That's why I had to add the `groups` migration file.

this seems to be caused by the following test:

```
tester.run_generate_migration(&vec![
    "CreateJoinTableUsersAndGroups",
    "count:int",
]);
```

kaplanelad:
From where the group came from?

floscodes:
> From where the group came from?

That also caught my attention 🙂

I didn’t intentionally introduce groups on my side.

From what I could tell, the groups table is implicitly required by the ActiveModel::delete implementation. When the table is missing, the delete call fails at runtime, which is what caused the test to break.

I traced this back to the CreateJoinTableUsersAndGroups migration generated in the test. That seems to be where the dependency comes from, even though groups itself isn’t created explicitly beforehand.

The migration was added purely to make the test pass and avoid the runtime error — happy to change this if there’s a more intended approach 👍

kaplanelad:
Please clarify  where the groups migration comes from specifically,  explain that ActiveModelTrait::delete requires all related tables to exist. this is not clear to me

 Also, please remove the commented-out test code in new.rs not sure if why it commended

floscodes:
Hi @kaplanelad, 

sorry for the delay.

It looks like the `groups` table may originate from one of the wizard tests. In `loco/loco-new/tests/wizard/new.rs`, the function `test_combination` calls:

```rust
tester.run_generate_migration(&vec!["CreateJoinTableUsersAndGroups", "count:int"]);
```

Inside `run_generate_migration`, the base command is constructed with:

```rust
let base_command = vec!["loco", "generate", "migration"];
```

and then extended with `CreateJoinTableUsersAndGroups`. The migration is executed via `self.run_migrate`, which during the test run creates a join table between `users` and `groups`.

From what I can see, creating a user in the wizard task works without issues. However, when the user is deleted, the presence of the `groups` table seems to become relevant. My assumption is that the wizard cannot find the corresponding migration in the base template, which may be why the deletion step fails. As a result, the wizard test appears to implicitly require the `groups` table to exist when deleting a user.

One possible explanation could also be related to how the ORM handles relations. Creating a user only inserts into the `users` table, so the join table might be not required at that stage. However, when deleting a user, the ORM may attempt to handle related records (for example in a `users_groups` join table). If those related tables do not exist, the deletion step could fail even though the insert worked correctly.

Please let me know if I might be misunderstanding something here.

Best regards!

jondot:
Hi @floscodes — thanks for sticking with this. `user:delete` alongside `user:create` is a genuinely nice addition and I'd like to land it, but not as-is for 1.0: adding a whole `groups` table to every generated app's schema to work around a test failure — without knowing why the test needed it — isn't something we can commit to at a 1.0 stability boundary. I think the real fix is on the wizard-test side (the `CreateJoinTableUsersAndGroups` step shouldn't create an implicit runtime dependency for an unrelated task test), not in the generated app. Could you check whether `ActiveModelTrait::delete` on `users::Model` actually touches a `groups` relation, or if the test is just bleeding state from a prior wizard step? Also, `run_test()`'s assertion got commented out in favor of a `println!` — that masks the failure, so let's not merge that. I'm going to defer this past 1.0 rather than adopt it now, but I'd love to bring `user:delete` in right after, once the `groups` mystery is actually resolved. Appreciate the persistence!

floscodes:
Hi, @jondot, thanks for the detailed feedback — you were right to push back, and I found the actual root cause.

`ActiveModelTrait::delete` on `users::Model` doesn't touch a `groups` relation at all (the generated entity has an empty `Relation` enum), and the `user:delete` task test doesn't reference `groups` either. The dependency was coming from somewhere else entirely: the wizard test already had a `CreateJoinTableUsersAndGroups` migration-generator step (predates this PR), which produces a migration with FK references to both `users` and `groups`. Since no app in this flow ever has a real `groups` table, `cargo loco db migrate` failed on that unrelated step — that's the error I was working around, not anything caused by `user:delete`.

I've pushed a fix that:
- Reverts the `groups` migration/table I'd added to `base_template` as a workaround (removed from `migration/src/lib.rs.t`, `setup.rhai`, and deleted `m20220101_000002_groups.rs`)
- Points that wizard-test step at `CreateJoinTableUsersAndMovies` instead, since `movies` already exists by that point in the flow (generated and auto-migrated a few steps earlier) — keeps the same coverage for the join-table naming inference without an implicit, unrelated dependency
- Restores the `run_test().expect(...)` assertion after project generation instead of the `println!` that was masking failures

So `user:delete` no longer touches the generated app's schema at all. Let me know if you'd like me to split this into a separate PR against the wizard test first, or if it's fine bundled here with `user:delete`/`user:create`.

jondot:
Rebased this onto post-1.1.0 master and opened it as #1809, with you as the commit author. Two small changes on the way in: the confirmation read moved to `spawn_blocking` (`stdin().read_line` inline in an `async fn` blocks the runtime thread, and this is template code people copy), and the `detail` string now shows its arguments the way `user:create` does.

I left out the wizard-test changes on your branch — uncommenting `test_all_combinations` and the join-table rename — since they're unrelated to the task itself. If you want those, they're worth their own PR and I'll look at them properly.

Sorry it sat this long. Thanks for the patience.

floscodes:
@jondot 

Thank you for picking this up and for the thoughtful improvements. I really appreciate your time and effort!

Issue #1294, opened 2025-03-10
Add MySQL as a Database Option


alokmahor:
## Feature Request

I was setting up a new project with `loco new` and noticed that the only database options available are **SQLite** and **PostgreSQL**—there's no MySQL option.

Since SeaORM (which Loco uses) supports MySQL, is there a specific reason why it's not included as a choice? 

**MySQL** is one of commonly used database system. If there is no good reason please add it as database option.

kaplanelad:
Hi @alokmahor,  
You’re right—since SeaORM supports MySQL, Loco can work with it too! To use MySQL, you can select `Postgres` in the CLI during `loco new`, then tweak the configuration: update the schema to `mysql://` in the config file and add `sqlx-mysql` as a feature to the `sea-orm` crate in your dependencies. After that, it should work smoothly.

Do you think adding it to the DB provider list in the CLI would be valuable?
```sh
❯ loco new
✔ ❯ App name? · myapp
✔ ❯ What would you like to build? · SaaS App with server-side rendering
? ❯ Select a DB Provider ›
Sqlite
Postgres
MySQL
None
```

alokmahor:
Hi @kaplanelad,

Yes, I understand that we can manually configure **MySQL** support by tweaking the configuration and dependencies, but my point is that since MySQL is a widely used database, it makes sense to include it as a first-class option in the CLI—just like **SQLite** and **PostgreSQL**.

For new users, having MySQL as a selectable option would make the setup much more intuitive. Right now, a newcomer might assume that MySQL isn't supported at all, since it's not listed as an option. Making it a part of the loco new flow would improve accessibility and reduce friction for those who prefer MySQL.

Additionally, SeaORM already supports MySQL, so there doesn’t seem to be a strong technical reason to exclude it. The extra steps required to set it up manually could be avoided by offering it as a built-in choice.

NexVeridian:
there are a [bunch of types](https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-structure/#column-type) that are different in mysql like unsigned, and some that are missing in mysql and sqlite like array

so maybe `loco-gen/src/mappings.json` should be changed to handle that later

alokmahor:
[...]

PR #1712, opened 2025-12-21
Improve MySQL compatibility with Loco Tests.

jmanucornejo:
I apologize in advance if I'm doing something wrong, this is one of my first times contributing to open source. All comments are appreciated and let me know if I can improve or have to do more for this to be considered.

Even though MySQL is not a standard Loco.rs database, there are users that run Loco with a MySQL Backend. It can be easily done adding some feature flags to seaorm.

But even with those flags, when running Loco.rs tests (for example in tests/models) and trying to seed fixtures to MySQL with  "seed::<App>(&boot.app_context).await.unwrap();", users will get this error returned  "Unsupported database backend: MySQL". 

I understand now it was because the tests go through specific functions in the db.rs which defaulted to errors on the  MySQL backend arm, making it impossible to test using MySQL backend.

I have implemented the MySQL arms for each of the functions that where causing errors. I also created some tests for MySQL backend for the functions which can be run with:

cargo test --features testing test_mysql

My 3 tests passed. I tried to replicate the postgres way of testing.

I also tried the formatting in the CONTRIBUTING.MD but it changed things I didn't contribute to so I tried my best to format it as I saw in other files.

### DIFF
--- a/src/tests_cfg/mod.rs
+++ b/src/tests_cfg/mod.rs
@@ -4,9 +4,9 @@ pub mod controllers;
 #[cfg(feature = "with-db")]
 pub mod db;
 #[cfg(test)]
-pub mod postgres;
-#[cfg(test)]
 pub mod mysql;
+#[cfg(test)]
+pub mod postgres;
 #[cfg(any(feature = "bg_pg", feature = "bg_sqlt"))]
 pub mod queue;
 #[cfg(test)]
--- a/src/tests_cfg/mysql.rs
+++ b/src/tests_cfg/mysql.rs
@@ -1,22 +1,28 @@
-use sqlx::MySqlPool;
 use std::time::Duration;
+
+use sqlx::MySqlPool;
 use testcontainers::{
-    ContainerAsync, GenericImage, ImageExt, core::{ContainerPort, WaitFor, logs::LogSource, wait::LogWaitStrategy}, runners::AsyncRunner
+    core::{logs::LogSource, wait::LogWaitStrategy, ContainerPort, WaitFor},
+    runners::AsyncRunner,
+    ContainerAsync, GenericImage, ImageExt,
 };
 
 /// Sets up a `MySQL` test container.
-/// 
+///
 /// # Returns
 ///
 /// A tuple containing the `MySQL` connection URL and the container instance.
 ///
 /// # Panics
 ///
-/// This function will panic if it fails to set up, start, or connect to the MySQL container.
+/// This function will panic if it fails to set up, start, or connect to the
+/// MySQL container.
 pub async fn setup_mysql_container() -> (String, ContainerAsync<GenericImage>) {
     let mysql_image = GenericImage::new("mysql", "8")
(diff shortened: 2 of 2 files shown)


kaplanelad:
Hey @jmanucornejo, thanks for opening this PR! 
Did you review this issue: https://github.com/loco-rs/loco/issues/1294 ?

jmanucornejo:
I did see that issue when starting my project. In fact, when I mentioned in my PR the feature flags, I meant the sqlx-mysql on sea-orm mentioned on that same issue. Everything seemed to work fine with that feature flag, **except tests**. 

Thats why I ended up doing this. I want to have MySQL for my tests in order for them to be consistent with my database backend. Apparently, the feature flag wasn't enough for tests. Seeding function kept throwing the "Unsupported database backend: MySQL".

kaplanelad:
I went over this and tested the scenario described in this comment:  
https://github.com/loco-rs/loco/issues/1294#issuecomment-2765840711

The issue appears to be related to the `generator` field. SeaORM generates different types for PostgreSQL and MySQL, which causes the mismatch.

I recommend running `cargo loco generate model ...` against a MySQL database to reproduce the issue and verify whether any errors occur.

jmanucornejo:
I'm not sure what that has to do with my pull request.  

To clarify, I have no problem generating models and entities against my MySQL Backend.

The problem is specifically with  **running tests**, and [ line 358, 400 and 452 in src/db.rs](https://github.com/loco-rs/loco/blob/master/src/db.rs):

` DatabaseBackend::MySql => {
      return Err(Error::Message(
          "Unsupported database backend: MySQL".to_string(),
      ));
  }`

The MySQL arm directly returns an error, even with MySQL working correctly. 

Can you please give me more detail as to what you want me to check about that comment?

kaplanelad:
Try generating a scaffold using the `cargo loco generate scaffold` command and add a variety of different db field types.  
Then verify that the generated code compiles successfully.

jmanucornejo:
As I mentioned before, I have no problem creating models, entities, etc using cargo loco commands. It does compile.

The problem is with tests.

Please I really don't see how what you are saying is related to the PR.

kaplanelad:
Ok, if everything works on your end, the next step is to implement all the MySQL cases in the project and add MySQL integration tests here:
https://github.com/loco-rs/loco/blob/master/loco-new/tests/wizard/new.rs

jmanucornejo:
Can you specify how that is related to this PR? I still don't understand why you keep mentioning the scaffolding or the issue #1294 . My PR is about running tests, not creating a new app.

If you need help implementing the app creatiion with MySQL, I can check it out, but this is not what this PR is about.

kaplanelad:
We do not officially support SQL right now. Adding official support would require integrating it into our tests and validating all related functionality.

jmanucornejo:
So in summary, if I want to get this to work, I'd need to create a PR with all mysql flows fixed, not just the tests part? Its an all or nothing kind of thing? Let me know so I can see what it would take. Thanks for your time!

Issue #1618, opened 2025-09-10
Task Command Link Arguments can't take file lists.


pweaver:
## Feature Request
I would like to use the task command line arguments in a more normal unixy way. In particular, I have some tasks which I want to grab a list of files like cargo loco task task_name *. The parser requires every argument to be like key:value which makes glob files and auto completion in the terminal non functional.

Preferably, there would be a way to pass all arguments (after the task name or a -- if thinks like the environment need to be separated) to the task as is to the task and we could write our own clap parser flexibly.

I'm happy to try implement and make a pull request but am not sure what design should happen for this.

I know I can just implement a separate binary, but that gets rid of the point of having a tasks api (setting up the environment for me).

kaplanelad:
Hey, 
can you share more? i not sure how how you expected to run it

pweaver:
Well, if I am writing a task to import a set of files I want to run
`cargo loco task import *` right now, it is impossible to specify multiple files, it throws a syntax error if you try this. (this is what I am trying to do right now)

Additionally, The colon syntax makes it very difficult to tab complete on linux. Even with a one parameter task, trying to write
cargo loco task import file:folder/na<tab> doesn't work because the file: part of the syntax breaks tab completion which makes it difficult to work with longer file names.

Having the only interface to tasks being this key/value pair makes it impossible to do anything complex with the command line api (see clap documentation, for example sub commands).

At minimum, having an interface which is a list of strings like a standard unix command would allow you to do most of this (you could for example just use clap to parse the rest of the args, or just use all args as a list of files).

kaplanelad:
We’d welcome a PR or prototype to experiment with this

PR #1730, opened 2026-01-15
Modify the Tasks API to take a [&String] instead of Vars.

pweaver:
This allows tasks to use Clap or just take a file list from the command line. Also create a VarsTask Trait to allow some backwards compatibility

This is a reference implementation for what I am thinking for #1618. It needs some work before it is ready to merge and I haven't fixed/written any tests of the documentation. Let me know your feedback on the design itself.

### DIFF
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -28,7 +28,7 @@ use crate::{
     prelude::BackgroundWorker,
     scheduler::{self, Scheduler},
     storage::{self, Storage},
-    task::{self, Tasks},
+    task::Tasks,
     Result,
 };
 
@@ -180,7 +180,7 @@ async fn shutdown_and_await_queue_worker(
 pub async fn run_task<H: Hooks>(
     app_context: &AppContext,
     task: Option<&String>,
-    vars: &task::Vars,
+    args: &[String],
 ) -> Result<()> {
     let mut tasks = Tasks::default();
     H::register_tasks(&mut tasks);
@@ -188,7 +188,7 @@ pub async fn run_task<H: Hooks>(
     if let Some(task) = task {
         let task_span = tracing::span!(tracing::Level::DEBUG, "task", task,);
         let _guard = task_span.enter();
-        tasks.run(app_context, task, vars).await?;
+        tasks.run(app_context, task, args).await?;
     } else {
         let list = tasks.list();
         for item in &list {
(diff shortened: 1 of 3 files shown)


jondot:
Sorry for the long silence — you asked for design feedback in January and deserved an answer well before now.

The problem is real and I want it fixed. `cargo loco task foo *.txt` is impossible today because the `value_parser` rejects anything that isn't `key:value`, so globs, file lists and shell completion are all off the table. Your CLI-layer change — `trailing_var_arg` collecting raw strings — is the part that actually unblocks that, and I'd take it as-is.

Where I'd push back is changing `Task::run` from `&Vars` to `&[String]`. That breaks every task anyone has written, plus the two that ship in the generated app template. For a capability that can be added without breaking anything, that's a steep price.

Counter-proposal: keep `Vars` as the task-facing type and give it the raw list.

```rust
pub struct Vars {
    pub cli: BTreeMap<String, String>,  // key:value pairs, parsed leniently
    args: Vec<String>,                  // every argument, verbatim
}

impl Vars {
    pub fn args(&self) -> &[String] { &self.args }
}
```

`Task::run` keeps its signature. `cli_arg` keeps returning `Result`. Existing tasks compile untouched. A task that wants clap feeds `vars.args()` straight into it, and a task that wants a file list just uses the slice — which is your original use case.

The lenient parse matters here: today an argument without a colon is a hard CLI error, so `*.txt` never reaches the task at all. Under this, non-`key:value` arguments simply don't appear in `cli` and are still present in `args()`.

One honest caveat: `Vars` isn't `#[non_exhaustive]` and `cli` is public, so anyone constructing `Vars { cli }` as a literal would break. That's rare, and worth sealing the type in the same change.

If that design works for you, I'd be glad to take a PR against it and target 1.2. If you think the raw-slice signature buys something this misses, say so — you've been living with the limitation and I haven't.

pweaver:
I can implement this. I am a little concerned about how the duel parsing would work from a security perspective. Any untrusted input could put colons in the filename/arguments which would populate the cli arg. we could put every argument in args but that would confuse someone using both.

jondot:
The colon collision is real, but I don't think it's a security boundary — argv comes from whoever can already run the binary, so nothing is being crossed. What's left is the ambiguity you're describing, and that's a docs problem.

On someone using both views: a task picks one. `args()` has to be complete and verbatim or clap can't parse it, which was the point of it. `cli` stays the convenience view for `key:value`.

Take the PR whenever you have time.

PR #1731, opened 2026-01-17
feat: `LocoOptionExt`: Helper methods to convert `Option<T>` into `loco_rs::Result<T>`

dsgallups:
## Objective
This PR adds new methods onto `Option<T>` to convert the None branch into an `Error`.

A common pattern I've found using loco is handling errors for model results when I need a specific model from sea_orm. For example, you usually have to do this in some controller:
```rust
let my_model = my_models::Entity::find_by_id(model_id)
  .await?
  .ok_or(Error::NotFound)?;
```
This is easy enough for converting into not found, but what if you'd like to do something else, like a custom error?
```rust
let my_model = my_models::Entity::find_by_id(model_id)
  .await?
  .ok_or(Error::CustomError(
      StatusCode::IM_A_TEAPOT,
      ErrorDetail::new("foo", "barbazqux")
  ));
```
After a while, this can get a big annoying to type everywhere. This PR adds new methods onto `Option<T>` to return errors with less syntax:
```rust
let my_model = my_models::Entity::find_by_id(model_id)
  .await?
  .status(
    StatusCode::IM_A_TEAPOT,
    "foo",
    "barbazqux"
  )?;

```

### LocoOptionExt::status
```rust
let optional_foo: Option<i32> = None;
let result: Result<i32> = optional_foo
    .status(StatusCode::BAD_REQUEST, "Missing number", "Maybe don't set optional_foo to None");

let Err(Error::CustomError(status, error_detail)) = result else {
    unreachable!();
};

assert_eq!(status, StatusCode::BAD_REQUEST);
assert_eq!(error_detail.error, Some("Missing number".to_string()));
assert_eq!(error_detail.description, Some("Maybe don't set optional_foo to None".to_string()));
```

### LocoOptionExt::msg
```rust
let optional_foo: Option<i32> = None;
let result: Result<i32> = optional_foo
    .msg("Where'd my number go?");
let Err(Error::Message(msg)) = result else {
    unreachable!();
};
assert_eq!(msg, "Where'd my number go?".to_string())
```

### LocoOptionExt::dbg()
```rust
let optional_foo: Option<i32> = None;
let result: Result<i32> = optional_foo.dbg();
let Err(Error::Message(msg)) = result else {
    unreachable!();
};

assert!(msg, "Found None::<i32> at src/errors.rs:7:40".to_string());
```

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -202,8 +202,9 @@ pub trait LocoOptionExt<T> {
     /// let Err(Error::Message(msg)) = result else {
     ///     unreachable!();
     /// };
-    ///
-    /// assert_eq!(msg, "Found None::<i32> at src/errors.rs:7:40".to_string());
+    /// // `msg` will return a message like
+    /// // "Found None::<i32> at src/errors.rs:7:40"
+    /// # assert!(msg.contains("Found None::<i32> at "));
     /// ```
     #[expect(clippy::missing_errors_doc)]
     #[cfg_attr(debug_assertions, track_caller)]


dsgallups:
Note: in order to not expose the `dbg` callsite and type information in release builds, I could wrap this information in an `#[cfg(profile = "dev")]` block

kaplanelad:
Thanks for the contribution! A few concerns:

 1. Value vs. complexity tradeoff: The existing patterns (`.ok_or(Error::NotFound)?, .ok_or_else(|| Error::CustomError(...))`) are already one-liners. Adding a new trait to the prelude for marginal savings increases the API
 surface all users need to learn.
 2. `dbg()` is gated behind `#[cfg(debug_assertions)]`, code using `.dbg()` will fail to compile in release mode. This is a footgun for users.

 Can you share a concrete before/after from a real controller that shows where this significantly improves readability over what we already have?

dsgallups:
You know what, I totally agree with you here. I think I'll keep these extension traits local to my repo. Thanks!

PR #1732, opened 2026-01-19
refactor: `Vars::cli_arg` returns `Result<&str>`

dsgallups:
This is a small QOL adjustment that allows you to quickly pattern match on a CLI argument. i.e.
```rust
match vars.cli_arg("foo")? {
  "apple" => ...
  "banana" => ...
  _ => ...
}
```
Sometimes when the value is optional...
```rust
match vars.cli_arg("foo") {
  Ok( "apple") => ...
  Ok(_) => ...
  Err(e) => ...
}
```

I don't see any particular behavior that would be lost from auto-dereffing the String [here](https://doc.rust-lang.org/std/string/struct.String.html). It seems that `&str` is more often more useful than `&String`.

### DIFF
--- a/loco-new/base_template/src/tasks/user_create.rs
+++ b/loco-new/base_template/src/tasks/user_create.rs
@@ -26,9 +26,9 @@ impl Task for UserCreate {
             .map_err(|_| Error::string("password is mandatory"))?;
 
         let register_params = RegisterParams {
-            email: email.clone(),
-            password: password.clone(),
-            name: name.clone(),
+            email: email.to_owned(),
+            password: password.to_owned(),
+            name: name.to_owned(),
         };
 
         // Create user with password using the same logic as register controller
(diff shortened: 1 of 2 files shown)


kaplanelad:
Hey @dsgallups, thanks for this! Nice ergonomic improvement for task args 👍
Minor change, **Use `.map(String::as_str)` instead of `.map(Deref::deref)`**, gives the same result without needing the `use std::ops::Deref` import,  and reads a bit more naturally.

dsgallups:
```
thread 'wizard::new::test_starter_combinations::case_3' (213569) panicked at tests/wizard/new.rs:72:10:
run clippy after create new project: Custom { kind: Other, error: "command [\"cargo\", \"clippy\", \"--\", \"-W\", \"clippy::pedantic\", \"-W\", \"clippy::nursery\", \"-W\", \"rust-2018-idioms\", \"-A\", \"clippy::result_large_err\"] exited with code 101" }
```
image
image

:thinking:

kaplanelad (approved):

jondot:
Thanks @dsgallups — `Vars::cli_arg` returning a `Result` is in 1.0.0. Closing in favor of the shipped version — appreciate the cleanup!

PR #1742, opened 2026-02-20
feat: Allow running scheduler and server without worker

### LINKED ISSUE #1737: Allow running scheduler and server without worker
## Feature Request

The current arguments to `start` only allow you to run the server or worker individually, the server and scheduler together, or all components together.  You cannot run the scheduler and server together without the worker.  I would think this could be a common pattern to scale workers horizontally while using a single server + scheduler to dispatch them work and the restriction doesn't seem intentional as far as I can tell.

Separation of CLI arguments to the `start` command to allow for all combinations of the components - `server`, `worker`, and `scheduler` - to be able to run together.

We've [implemented](https://github.com/Adversarial-Risk-Management/loco/pull/1) this with individual CLI arguments for `start`, deprecating the `--server-and-worker` arg. Since there is also a `scheduler` subcommand, perhaps a more holistic change could address that as well.

mccormickt:
Separates the CLI arguments into --server, --worker, and --scheduler to allow components to be run in all combinations. Retains --server-and-worker and --all for backwards compatibility

Fixes #1737

### DIFF
--- a/src/boot.rs
+++ b/src/boot.rs
@@ -40,8 +40,16 @@ pub enum StartMode {
     ServerOnly,
     /// Run the application web server and the worker in the same process.
     ServerAndWorker,
+    /// Run the server and scheduler without workers.
+    ServerAndScheduler,
     /// Pulling job worker and execute them
     WorkerOnly {
+        /// Specifies that the worker should only handle jobs associated with one of these tags.
+        /// If empty, the worker handles all jobs.
+        tags: Vec<String>,
+    },
+    /// Run workers and scheduler without the HTTP server.
+    WorkerAndScheduler {
         /// Specifies that the worker should only handle jobs associated with one of these tags.
         /// If empty, the worker handles all jobs.
         tags: Vec<String>,
@@ -462,6 +470,15 @@ pub async fn run_app<H: Hooks>(mode: &StartMode, app_context: AppContext) -> Res
                 run_scheduler: false,
             })
         }
+        StartMode::ServerAndScheduler => {
+            let router = setup_routes::<H>(&app_context, &initializers).await?;
+            Ok(BootResult {
+                app_context,
+                router: Some(router),
+                worker: None,
+                run_scheduler: true,
+            })
+        }
         StartMode::All => {
             register_workers::<H>(&app_context).await?;
             let router = setup_routes::<H>(&app_context, &initializers).await?;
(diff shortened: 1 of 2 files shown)


jtwaleson:
I looked into this code before but was a bit puzzled, would it not be better to not have one value for the combination of three "jobs"? Maybe `start --worker --server --scheduler` is better than `start --all`.

mccormickt:
> I looked into this code before but was a bit puzzled, would it not be better to not have one value for the combination of three "jobs"? Maybe `start --worker --server --scheduler` is better than `start --all`.

The intent was to not break folks that are using `--all` by removing the option. I would agree that supporting only the separate args would be better longer term.

kaplanelad:
Thanks, the ability to run scheduler + server without worker is a real need and I appreciate you tackling it.

I'd like to suggest a simpler approach that avoids the breaking change to `StartMode`. Since `StartMode` is public API, changing it from an enum to a struct will break downstream code that matches on it.

Instead, we can solve this by just adding a `--scheduler` flag and two new enum variants:

```rust
pub enum StartMode {
    ServerOnly,
    ServerAndWorker,
    ServerAndScheduler,                        // new
    WorkerOnly { tags: Vec<String> },
    WorkerAndScheduler { tags: Vec<String> },  // new
    All,
}
```

The CLI usage would be:
- `cargo loco start --scheduler` → server + scheduler (default is server, so `--scheduler` adds to it)
- `cargo loco start --worker --scheduler` → worker + scheduler
- `cargo loco start --server-and-worker --scheduler` → maps to `All`

Everything else stays the same, no struct redesign, no deprecations, no breaking changes. This covers all the combinations from the original issue.

Also, a heads-up on the current PR: `start()` in `boot.rs` binds `run_scheduler` as `_` and never acts on it, so `--scheduler` wouldn't actually start the scheduler as-is. And `cargo loco start --scheduler` alone hits the `(None, None) => {}` catch-all and silently exits.

mccormickt:
> The CLI usage would be:
> 
> * `cargo loco start --scheduler` → server + scheduler (default is server, so `--scheduler` adds to it)
> * `cargo loco start --worker --scheduler` → worker + scheduler
> * `cargo loco start --server-and-worker --scheduler` → maps to `All`

IMO, the coupling of server to worker -- but not scheduler -- is a bit counter-intuitive.  I'm happy to make the requested edits to keep from introducing breaking changes, however.

mccormickt:
@kaplanelad  Updated with https://github.com/loco-rs/loco/pull/1742/commits/b1af0c50ec959b84bb90b8cb50c9fc98133f7d4d

mccormickt:
@kaplanelad Friendly bump on this when you get the chance 🙏

jondot:
Thanks @mccormickt — running the scheduler without a worker landed in 1.0.0. Closing in favor of the shipped version, which grew out of this. Appreciate you scoping it!

PR #1752, opened 2026-03-09
fix(sec): update bytes and jsonwebtoken

AnthonyMichaelTDM:
Updates jsonwebtoken to a version that fixes [CVE-2026-25537](https://github.com/advisories/GHSA-h395-gr6q-cpjc) (authentication bypass via type confusion)

- I don't have a poc demonstrating loco is vulnerable to this, but the POC in the gh security advisory (linked) seems similar to how jsonwebtoken is used in loco so it's better safe than sorry imo

Updates bytes to a version that fixes [CVE-2026-25541](https://github.com/advisories/GHSA-434x-w66g-qw3r) (integrer overflow)


AnthonyMichaelTDM:
idk, what do you think is the best way to introduce these features that'll work within the current test suite?

I'd like to avoid just choosing one option and not making it configurable through feature flags

esheri3:
Why does updating jsonwebtoken and bytes require the introduction of two new dependencies? Those two new deps seem to be the reason for a possible feature flag.

AnthonyMichaelTDM:
What new dependencies?

esheri3:
I believe these two if I read the changes correctly:

```
jwt_aws_lc_rs = ["dep:jsonwebtoken", "jsonwebtoken/aws_lc_rs"]
jwt_rustcrypto = ["dep:jsonwebtoken", "jsonwebtoken/rust_crypto"]
```

In my opinion, getting these crates updates (esp jsonwebtoken) is very important for the security of loco.rs framework. If there's friction to adopting this fix because of the additional feature, perhaps split it into two PRs, one just updates jsonwebtoken and bytes as-is (no new features), and the second introduces the feature w/ default ensuring use of aws_lc_rs?

Just trying to help support this discussion, as I value these proposed changes.

AnthonyMichaelTDM:
Those are feature flags used to specify which crypto engine you want to use

jondot:
Thanks @AnthonyMichaelTDM — and for erring on the side of caution here. Good news: 1.0.0 already ships past both advisories:

- `jsonwebtoken = 10.3.0` — CVE-2026-25537 / GHSA-h395-gr6q-cpjc was fixed in 9.3.1 and carried through the 10.x line, and `auth` uses `jsonwebtoken/rust_crypto`.
- `bytes = 1.11` — clears GHSA-434x-w66g-qw3r.

So the current release isn't exposed to either, and there's no separate patch needed. Closing since it's covered in 1.0.0 — genuinely appreciate you tracking the CVEs and flagging them.

Issue #1749, opened 2026-03-07
Newly generated project doesn't start


Leandros:
**Description**

Creating a new loco project fails to build the first time.

**To Reproduce**

```
$ cargo install loco
$ loco new
$ cargo loco start
[...]
Error: Message("Failed to add FTL resources to the bundle")
```

**Expected Behavior**

Newly generated project should start

**Environment:**

loco v0.16.4

rustc 1.90.0 (1159e78c4 2025-09-14)

m1gh:
I ran into the same issue, because shared.ftl was inside the I18N_DIR directory, the loader reads it during the directory scan, and then read it again because it was explicitly specified in .shared_resources(). I fixed this by moving shared.ftl outside of the i18n directory so it isn't scanned twice.

I moved assets/i18n/shared.ftl to assets/shared.ftl.
I updated the path in src/initializers/view_engine.rs to point to the new location: const I18N_SHARED: &str = "assets/shared.ftl";

the app now successfully runs.

jondot:
Fixed for 1.0.0 — thanks @Leandros for reporting and @zjom / @lunfel for the fix (#1758). Root cause: `shared.ftl` living inside `assets/i18n/` was picked up by the locale scan *and* `.shared_resources()`, which newer fluent-templates rejects as a duplicate. Fixed by relocating it to `assets/shared.ftl`, shipped in 1.0.0. Closing.

PR #1758, opened 2026-04-16
new project generator fix (#1749)

zjom:
Fix issue where new project generated with `loco new` does not start due to shared.ftl being loaded twice. Once during scan of `assets/i18n/` and again when specified by `.shared_resources` in builder.

CHANGES
Moved `assets/i18n/shared.ftl` to `assets/shared.ftl` and updated `I18N_SHARED` constant to point to new path.


lunfel:
Hey @zjom! I was looking into why this issue wasn't caught by the CI before making it to master. It turns out that running cargo test on the generated project actually flags the errors. I went ahead and added that step to our workflow.

Would you mind including my changes in your PR so we have a regression test for this moving forward?

https://github.com/lunfel/loco/tree/master

You can see [here](https://github.com/lunfel/loco/actions/workflows/loco-rs-ci-sanity.yml), the [first run](https://github.com/lunfel/loco/actions/runs/24807071409) was without the fix and the workflow fails, then I merged in your fix and then the [second run](https://github.com/lunfel/loco/actions/runs/24807622246) passes

zjom:
@lunfel Amazing :)

lunfel:
@zjom I missed a test that was failing. I fixed in my last commit: https://github.com/loco-rs/loco/commit/44551a9d172d3b1b796f6fff2ab48fc598d0bfeb. Can you include it? Everything should be green with this fix

D-system:
The file [`docs-site/content/docs/the-app/views.md` ](https://github.com/loco-rs/loco/blob/e9598a6822c581a06416bb668ad80e1920e54d09/docs-site/content/docs/the-app/views.md#L102) should be updated.

jondot:
Thanks @zjom — and @lunfel for the regression-test gap and the fix! This is already resolved in 1.0.0 ("keep shared i18n resource outside the scanned locale dir"): `shared.ftl` moved to `assets/shared.ftl` so the locale scanner never double-registers it — the same approach you took. I also folded in your CI-hardening idea (`cargo test --all-features` now runs against the generated app), and credited you both as co-authors. Closing in favor of that fix since 1.0.0 supersedes it — really appreciate the diagnosis.

PR #1762, opened 2026-04-27
docs: migrate 8 README files to a single NRG template

andriishin:
## Why

`README.md`, `README.es.md`, `README.fr.md`, `README.ja.md`, `README.ko.md`, `README.ru.md`, `README-pt_BR.md`, `README-zh_CN.md` — 8 hand-maintained README files. With this many translations, every README change means editing the same paragraph in 8 places.

This PR migrates all eight files to a single source-of-truth template using [Nanolaba Readme Generator (NRG)](https://github.com/nanolaba/readme-generator). Combined with [`nanolaba/nrg-action@v1`](https://github.com/nanolaba/nrg-action), regeneration happens in CI on every push to `master` — no toolchain locally, just YAML.

## What changes

- **New:** `README.src.md` — the unified source. Each line is tagged with `<!--<lang>-->` so it appears only in that language's output. Generated programmatically; outputs are byte-identical to today's files (modulo a 2-line auto-generated header).
- **New:** `.github/workflows/nrg.yml` — runs `nanolaba/nrg-action@v1` (pinned to NRG `1.1`):
  - Push to `master` touching `README.src.md` or workflow → regenerates and auto-commits.
  - PR touching any `README*.md` or template → drift check (`mode: check`) — fails with unified diff when a contributor edits a generated file directly.
- **Modified:** all 8 README files — regenerated. The only change is the 2-line auto-generated header at the top.

The two non-default-pattern filenames (`README-pt_BR.md` and `README-zh_CN.md`, which use a hyphen separator instead of `.lang.md`) are preserved through `<!--@nrg.fileNamePattern.pt_BR=README-pt_BR.md-->` and the equivalent for `zh_CN` — NRG's other 5 translations use the standard `.lang.md` pattern as before.

## Verification

Locally, regenerating with NRG `1.1` from Maven Central against current `master` (commit e9598a6) produces output that diffs against the originals only by the 2-line auto-generated header. Zero functional content change.

## What's intentionally NOT in this PR

- **Translation alignment.** Today's content lives unchanged in `README.src.md`. Future PRs can extract shared structural elements (badges, install snippets) into shared regions of the template.
- Nothing in the Rust crates/build is touched.

## How to regenerate locally

```bash
curl -L -o nrg.jar https://repo1.maven.org/maven2/com/nanolaba/readme-generator/1.1/readme-generator-1.1.jar
java -jar nrg.jar -f README.src.md
```

## About NRG

[`com.nanolaba:readme-generator`](https://central.sonatype.com/artifact/com.nanolaba/readme-generator) — open-source, Apache 2.0, Java 8+. Ships as CLI, Maven plugin, GitHub Action, and library. Zero local toolchain needed when used through the action.


D-system:
The new file has everything in it. I don't see the value in it. I personally find that harder to understand as I need to read markdown, the newly `<!--xx-->` sprinkle in the file and find the bit that I would like to update.


Note: I'm not a maintainer.

jondot:
Thanks @andriishin — this is genuinely nice work, and a single-source README template is the right instinct for keeping 8 languages in sync. For right now, though, we're going to keep the per-language READMEs hand-maintained rather than add a generator + CI workflow to the release surface so soon after 1.0 — I'd rather not introduce new tooling into that path while things are settling. So I'm closing this as declined-for-now, with real appreciation — if the hand-maintained approach starts to bite, this is exactly where we'll come back to. 🙏

PR #1764, opened 2026-05-03
feat(mailer): add MultiEmail and mail_template_multi for multiple recipients

YtimoDeng:
## Problem

The current `Mailer` API only supports a single recipient in the `to`, `cc`, and `bcc` fields (`String`). There is no built-in way to send one email to multiple recipients in a single call.

This is a common real-world need — notification emails, team digests, CC-to-manager flows — and currently requires users to loop and call `mail_template` once per recipient, which is inefficient and semantically incorrect (each call produces a separate email rather than one email with multiple recipients).

## Solution

Introduces a parallel set of types alongside the existing single-recipient API. The existing `Email`, `Args`, `MailerWorker`, and `mail_template` are **completely untouched** — no breaking changes.

### New additions

| Symbol | Description |
|---|---|
| `MultiEmail` | Serializable job payload with `to`, `cc`, `bcc` as `Vec<String>` |
| `MultiArgs` | Caller-facing args struct with `to`, `cc`, `bcc` as `Vec<String>` |
| `MultiMailerWorker` | Background worker that processes `MultiEmail` via the existing `mailer` queue |
| `Mailer::mail_multi` | Dispatches a `MultiEmail` to `MultiMailerWorker` |
| `Mailer::mail_template_multi` | Renders a template and calls `mail_multi` |
| `EmailSender::mail_multi` | Builds and sends a lettre message from `MultiEmail` |

### Usage

```rust
MyMailer::mail_template_multi(
    &ctx,
    &TEMPLATES,
    MultiArgs {
        to: vec!["alice@example.com".to_string(), "bob@example.com".to_string()],
        cc: vec!["manager@example.com".to_string()],
        locals: json!({ "name": "Team" }),
        ..Default::default()
    },
)
.await?;
```

## Testing

- Existing snapshot tests (`can_send_email`, `can_send_email_with_custom_headers`) pass unchanged, confirming no regression on the original path.
- Three new snapshot tests cover `mail_multi` with multiple `to`, `cc`, and `bcc` recipients, verifying the correct headers appear in the built message.
- `cargo fmt` (nightly) — no changes required.
- `cargo clippy -p loco-rs` — zero warnings in mailer code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/mailer/email_sender.rs
+++ b/src/mailer/email_sender.rs
@@ -9,7 +9,7 @@ use lettre::{
 };
 use tracing::error;
 
-use super::{Email, Result, DEFAULT_FROM_SENDER};
+use super::{Email, MultiEmail, Result, DEFAULT_FROM_SENDER};
 use crate::{config, errors::Error};
 
 /// An enumeration representing the possible transport methods for sending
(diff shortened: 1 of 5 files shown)


jondot:
Thanks @YtimoDeng — multi-recipient email (multiple to/cc/bcc) shipped in 1.0.0, which is what `MultiEmail` / `mail_template_multi` were after. Closing in favor of the 1.0.0 mailer work. Thank you!

PR #1771, opened 2026-06-09
Apply auto-formatting

D-system:
Applying only the auto-formatting and avoid having them in unrelated PRs (eg: https://github.com/loco-rs/loco/pull/1758/commits).

Use command from the [documentation](https://github.com/loco-rs/loco/blob/e9598a6822c581a06416bb668ad80e1920e54d09/CONTRIBUTING.md#code-style).

### DIFF
--- a/loco-new/src/generator/template.rs
+++ b/loco-new/src/generator/template.rs
@@ -51,7 +51,9 @@ impl Template {
     #[must_use]
     pub fn is_template(&self, path: &Path) -> bool {
         path.extension()
-            .and_then(|ext| ext.to_str()).as_ref().is_some_and(|&ext| ext == TEMPLATE_EXTENSION)
+            .and_then(|ext| ext.to_str())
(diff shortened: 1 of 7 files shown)


jondot:
Thanks @D-system — appreciate isolating the formatting/clippy cleanup out of feature PRs. Against 1.0.0 it's a mixed bag after the 1.0 generator rebuild: the `schema.rs` change is already superseded (the rebuild restructured that enum match the same way and further), and a couple of the `cli.rs` one-liners are already on the branch while the `port.map_or → unwrap_or` fix now needs to land in two places. I've pulled the still-applicable pieces in with credit — the loco-new `is_some_and`, and the xtask `.clone()`/`const fn`/`#[must_use]`/stray-`;` fixes. Closing since the remainder is overtaken by the rebuild — thanks again for keeping formatting out of feature PRs, it makes reviews much easier.

PR #1774, opened 2026-06-22
feat(mailer): support implicit TLS (SMTPS / port 465)

### LINKED ISSUE #1773: SMTP mailer can't use implicit TLS (port 465 / SMTPS) — secure: true only does STARTTLS
## Description

The SMTP mailer only supports STARTTLS. `EmailSender::smtp()` builds the transport as `starttls_relay` when `secure: true`, otherwise `builder_dangerous` (cleartext). There is no way to use **implicit TLS** (SMTPS, port 465), where the socket is encrypted from the first byte.

Providers such as Mailgun and Gmail offer/recommend implicit TLS on **465**. Pointing the current config at 465 with `secure: true` produces a broken STARTTLS-on-465 connection that fails — and **silently**, because the send runs in the mailer worker: the request returns fine and the email simply never leaves. (Discussion #597 even shows `port: 465, secure: true` as an example, which cannot work today.)

## Repro

```yaml
mailer:
  smtp:
    enable: true
    host: smtp.eu.mailgun.org
    port: 465
    secure: true
    auth: { user: postmaster@mg.example.com, password: ... }
```
→ mail never sends; no surfaced error.

## Expected

Be able to select implicit TLS for port 465.

## Proposed fix

A backward-compatible `tls: starttls | implicit | none` option on `SmtpMailer`. When set it selects the transport; when unset, fall back to the legacy `secure` flag (`true` → starttls, `false` → none) so existing configs are unaffected. `lettre` already exposes `relay()` (implicit TLS) next to `starttls_relay()`, so the change is small.

PR ready.

zmilan:
## What

Adds implicit-TLS support to the SMTP mailer so it can reach providers that listen for TLS on port 465 (SMTPS). Today `secure: true` only ever does STARTTLS (`starttls_relay`), so 465 cannot work and fails silently in the mailer worker.

Fixes #1773.

## How

- New `MailerTls` enum (`starttls` | `implicit` | `none`) and an optional `SmtpMailer.tls` field.
- `SmtpMailer::tls_mode()` resolves the effective mode: explicit `tls` wins; otherwise it is derived from the legacy `secure` bool (`true` → starttls, `false` → none).
- `EmailSender::smtp()` matches the mode → `starttls_relay` / `relay` (implicit) / `builder_dangerous`.

## Backwards compatibility

Fully compatible — `tls` is `#[serde(default)]` optional; configs that only set `secure` behave exactly as before.

## Config

```yaml
mailer:
  smtp:
    host: smtp.example.com
    port: 465
    tls: implicit
```

## Tests

Added `smtp_builds_for_each_tls_mode` covering all three modes, the legacy `secure` mapping, and `tls`-over-`secure` precedence. `cargo test --lib mailer` green; the two clippy warnings in CI are pre-existing (`schema.rs`, `engine.rs`), not in the touched files.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -529,6 +529,20 @@ pub enum WorkerMode {
 ///     port: 1025
 ///     secure: false
 /// ```
+///
+/// Example (production) with implicit TLS on port 465 (SMTPS) — note `tls: implicit`,
+/// since `secure: true` only does STARTTLS (port 587):
+/// ```yaml
+/// mailer:
+///   smtp:
+///     enable: true
+///     host: smtp.example.com
+///     port: 465
+///     tls: implicit
+///     auth:
+///       user: postmaster@mg.example.com
+///       password: "{{ get_env(name='SMTP_PASSWORD') }}"
+/// ```
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Mailer {
     pub smtp: Option<SmtpMailer>,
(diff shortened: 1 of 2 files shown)


zmilan:
Closing — this change has already landed on \`master\` via #1783 (Loco 1.0.0), which lists "implicit TLS (465)" among the mailer changes.

The 1.0.0 squash split \`src/config.rs\` into \`src/config/*\`, which is what made this PR show conflicts. Comparing the two: \`src/config/mailer.rs\` on \`master\` now carries the \`MailerTls\` enum (\`starttls\` | \`implicit\` | \`none\`), the \`#[serde(default)] tls: Option<MailerTls>\` field, and \`SmtpMailer::tls_mode()\`; \`EmailSender::smtp()\` matches on \`tls_mode()\` → \`starttls_relay\` / \`relay\` / \`builder_dangerous\`; and the \`smtp_builds_for_each_tls_mode\` test is in place. Nothing here is left to merge, so rebasing would just produce an empty commit.

Closing this and #1773 as resolved by 1.0.0.

PR #1778, opened 2026-07-19
fix(docs): flash of light theme in dark mode

david-szabo97:
Fixes https://github.com/loco-rs/loco/issues/1333

The JS code responsible for the dark mode class is now inlined, so it runs before the page is rendered. No more flashbang in dark mode.

PR #1779, opened 2026-07-19
fix(docs): set images width and height to avoid content jumping

david-szabo97:
Missing `width` and `height` attributes caused content jumping on several pages. Fixed by adding the attributes to all image elements.


jondot:
Thanks @david-szabo97 — same note as #1778: the new 1.0 site (Astro/Starlight) supersedes the old docs-site this patches (image dimensions / layout shift are handled there). Closing with thanks — appreciate the attention to detail.

PR #1798, opened 2026-08-04
chore(deps): bump opendal to 0.58.1

Xuanwo:
## Why

Bump the direct `opendal` dependency from `0.57` to `0.58.1` so Loco's storage backends stay on a current OpenDAL release.

## Changes

- Pin `opendal` to `0.58.1` in the root `Cargo.toml` (same feature set: memory/fs/retry + optional s3/azure/gcs).
- Adapt storage drivers to OpenDAL 0.58 public API:
  - `Operator::new(...)` now returns a finished `Operator` (remove `.finish()`).
  - Replace `info().full_capability()` with `info().capability()`.

## Validation

- `cargo check -p loco-rs --features all_storage` (resolves `opendal` **0.58.1**).
- `cargo test -p loco-rs --lib storage --features "all_storage,testing,worker_redis"` — 40 storage tests passed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,7 +140,7 @@ chrono = { workspace = true }
 uuid = { version = "1.10.0", features = ["v4", "fast-rng"] }
 
 # File Upload
-opendal = { version = "0.57", default-features = false, features = [
+opendal = { version = "0.58.1", default-features = false, features = [
     "services-memory",
     "services-fs",
     "layers-retry",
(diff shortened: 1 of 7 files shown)


jondot:
Shipped in 1.1.0 — `opendal` is on 0.58.1 in the tree now. Thanks, and you're credited on the release commit.

Related: the storage how-to used to tell people to pin `opendal = "0.57"` themselves, which would have fought this. That text is gone.

Closing — https://github.com/loco-rs/loco/releases/tag/v1.1.0

PR #1802, opened 2026-08-07
fix(testing): redact timestamps with negative UTC offsets

onyxblade:
## The bug

`testing::redaction::get_cleanup_date` returns three ordered filters. The first one — the only rule that consumes a UTC offset — ends in `\+\d{2}:\d{2}`, so it matches a **positive** offset only:

```rust
(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?\+\d{2}:\d{2}", "DATE"), // with tz
(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+", "DATE"),
(r"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})", "DATE"),
```

`DateTimeWithTimeZone` debug-prints as RFC 3339 with an explicit numeric offset whose sign follows the local timezone. On a machine **west of UTC** the first rule misses, the timestamp falls through to rules 2/3, and those stop at the seconds — leaving the offset stranded:

```
-            DATE,
+            DATE-03:00,
```

Every snapshot carrying a timestamp then fails. CI never caught it because GitHub runners are UTC, where the offset is `+00:00` and rule 1 matches.

## Reproduction

On a host with a negative-offset `TZ` (reproduced on `America/Halifax`, UTC−3, and confirmed to pass under `TZ=UTC`):

```sh
loco new --name hello_loco --db sqlite --bg async --assets none
cd hello_loco && cargo test
```

```
failures:
    requests::auth::can_register
    requests::auth::can_resend_verification_email

test result: FAILED. 24 passed; 2 failed
```

Both failures are the `email_verification_sent_at` field redacting to `DATE-03:00`. So a freshly generated app has a red test suite out of the box for anyone west of UTC — the pattern has been this way since #1047 (2024-12-03).

## The fix

Widen the sign to `[+-]`. One character; no behavior change on UTC or positive offsets, since `[+-]` is a strict superset of `\+`. No existing snapshot needs regenerating.

## Notes for the reviewer

- I deliberately left the `Z` suffix form unhandled — `DateTimeWithTimeZone` is `chrono::DateTime<FixedOffset>`, whose `Debug` always renders a numeric offset (`+00:00`), never `Z`, so it isn't reachable from these snapshots. Happy to widen it if you'd rather the filter be format-complete.
- `old_date_regex_only_matched_positive_offsets` follows the style of the existing `old_password_regex_was_inert` guard: it pins the old pattern's behavior so the regression can't quietly return.

## Testing

- `cargo test --all-features --lib testing::redaction` → 5 passed, including the two new tests.
- `cargo test --all-features --lib` → 494 passed, 0 failed.
- `cargo fmt --all -- --check` → clean.
- `cargo clippy --all-features -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms` → clean.

### DIFF
--- a/src/testing/redaction.rs
+++ b/src/testing/redaction.rs
@@ -22,7 +22,7 @@ pub fn get_cleanup_date() -> &'static Vec<(&'static str, &'static str)> {
     CLEANUP_DATE.get_or_init(|| {
         vec![
             (
-                r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?\+\d{2}:\d{2}",
+                r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?[+-]\d{2}:\d{2}",
                 "DATE",
             ), // with tz
             (r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+", "DATE"),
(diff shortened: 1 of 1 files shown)


jondot:
Shipped in 1.1.0 — thanks for catching this one.

It was worse than it looked: a freshly generated app had a red test suite out of the box for everyone west of UTC, and had since 0.14. CI never noticed because GitHub runners are UTC.

Apologies that your name didn't make it onto the co-author list of the release commit — an oversight on my side while assembling it, not a judgment on the fix. The changelog entry credits this PR.

Closing since it's in — https://github.com/loco-rs/loco/releases/tag/v1.1.0

PR #1805, opened 2026-08-09
storage: add list/stat/exists to storage instance api

robertazzopardi:
## Summary

Adds `exists` / `list` / `stat` on the `Storage` instance api. I've tried to follow the existing read write apis closely.

`StoreDriver` and `StorageStrategy` now require those methods, which is a breaking change.

Mirror mode in `ReplicatedStrategy` falls back on miss/empty/error, skips empty secondary lists so they don't hide data (or swallow a primary error). Backup stays primary-only.

## Validation

In a fresh loco app, added mem storage to `AppContext` in `after_context`, added a small `/api/storage` controller, then hit it with curl / request tests: upload → exists → list (recursive + not) → stat → delete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 
+## Unreleased
 
+### Breaking Changes
+
+- **Storage `StoreDriver` / `StorageStrategy` gain required `list` and `stat`
+  methods** (and `StorageStrategy` also requires `exists`). Custom driver and
+  strategy implementations must implement them. Built-in drivers/strategies
+  already do. The `Storage` facade exposes matching `exists` / `list` / `stat`
+  (and `*_with_policy`) APIs alongside the existing upload/download/delete
+  surface.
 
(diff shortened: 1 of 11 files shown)


robertazzopardi:
Decided to fix the sanity check clippy lints in a commit on this pr, despite not being related to the changes. Happy to do that in a separate pr in a more dedicated fix.

jondot:
Shipped in 1.1.0 — thanks, this was a real gap in the storage API.

It grew a bit on the way in: `list`/`stat` became required on `StoreDriver` and `StorageStrategy` (plus `exists` on the strategy), and the facade got the `*_with_policy` variants alongside. Under `ReplicatedStrategy` in mirror mode, `exists` and `list` now treat a miss the same way the other reads treat an error and fall through to secondaries; backup mode stays primary-only.

You're credited on the release commit. Closing this since it's in — https://github.com/loco-rs/loco/releases/tag/v1.1.0

PR #1783, opened 2026-07-29
Loco 1.0.0 — first stable release

jondot:
Loco 1.0.0 — the first stable release. A single, intentionally-breaking milestone.

**Headline:** move to **Sea-ORM 2.0** (with sqlx 0.9). For most apps the upgrade is mechanical — the `schema` helpers and generated model/migration shapes absorb the API churn. Full step-by-step guide in the upgrade docs.

**Also in 1.0:**
- First-class support for coding agents — every new app ships `AGENTS.md`; the site serves `llms.txt` / `llms-full.txt`.
- Background jobs: priority queues, an opt-in reaper, `perform_later` returns the job id, pluggable `QueueProvider`, scheduler-without-worker.
- Mailer: multi-recipient, Tera template inheritance, implicit TLS (465), custom headers, synchronous `deliver_now`.
- Web: optional `Option<JWT>` extraction, verb-explicit route builders, `MiddlewareStackExt` for surgical stack edits.
- Ops: TLS to managed Postgres/Redis, typed streaming `db::dump`.
- A security & correctness sweep: local storage no longer roots at `/`, honest 4xx/5xx error mapping, rebuilt `remote_ip` trust model, HMAC-only JWT, 64-bit primary keys by default.

MSRV is **1.94** (sea-orm 2.0.0 declares it).

Full detail in `CHANGELOG.md`.

Issue #1727, opened 2026-01-12
development.yml fails when there are YAML evals + long lines + prettier + format_on_save


joelparkerhenderson:
Loco uses the file `config/development.yml` and similar ones, but these are failing due to glitches with some editors and formatters, notably zed, prettier, and the zed "format_on_save" setting.

Example file line:

```yaml
uri: {{ get_env(name="DB_URL", default="postgres://username:password@hostname:5432/database") }}
```

To reproduce the failure, open the file in the zed editor with zed default settings such as a fresh install. Make any edit, then save. This causes zed + prettier to do an automatic reformat that forces multi-line and inserts commas, like this:

```yaml
uri:
  {
    { 
      get_env(name="DB_URL", default="postgres://username:password@hostname:5432/database") 
    },
  }
```

This glitch causes Loco to fail on start.

I've reported the problem here: https://github.com/zed-industries/zed/issues/46632

The root cause seems to be that the Loco configuration file format is sort-of-YAML but sort-of-not YAML. I'm generally aware of these kinds of YAML glitches in other tools, such as with Ansible and also templating syntaxes.

My workaround is to add a zed settings file that disables the global "*_on_save" settings, and later I'll add specific syntax adjustments for specific languages and file names:

```json
{
  "format_on_save": false,
  "remove_trailing_whitespace_on_save": false,
  "ensure_final_newline_on_save": false
}
```

Could Loco eventually upgrade to using Rust instead of YAML for the configuration files? IMHO upgrading from YAML to Rust would be safer, and also easier to edit, and also more like Rails which uses Ruby for its config files.

Thanks for your consideration, and for Loco.

afarinetti:
It seems to be the `{{` getting reformatted to `{ {` that is doing the above for me.

I had to disable the `yaml-language-server` formatting in Zed to get it to stop doing it, for me. Obviously, this is non-ideal.

My project `.zed/settings.json` file:

```json
{
    "languages": {
        "YAML": {
            "formatter": "language_server"
        }
    },
    "lsp": {
        "yaml-language-server": {
            "settings": {
                "yaml": {
                    "format": {
                        "enable": false
                    }
                }
            }
        }
    }
}
```

arferreira:
I suffered a lot @joelparkerhenderson with this problem on my nvim, but as @afarinetti mentioned above, that LSP setting will help you to fix that easily.

joelparkerhenderson:
>  that LSP setting will help you to fix that easily.

[...]

PR #1699, opened 2025-11-26
Add AWS Lambda deployment support (Zappa-like experience)

SMCodesP:
## Summary

This PR adds native AWS Lambda deployment support to Loco, aiming to provide a simple way to deploy applications to AWS Lambda with just a single command—much like [Zappa](https://github.com/zappa/Zappa) does for Python projects. It's my first Pull Request here, so I'm really excited but also a bit nervous; I'm not entirely sure if I should be opening this yet or if it's ready for review. If there's anything I need to adjust, add, or fix, please just let me know—I'm happy to iterate!

## Motivation

While Loco already supports Docker and Shuttle deployments, AWS Lambda support opens the door to the massive AWS ecosystem.

The implementation is inspired by **Zappa**, the popular Python framework for serverless deployment. Like Zappa, this feature:
- Requires minimal configuration
- Creates temporary files during deployment and cleans up automatically
- Provides a simple CLI interface for deploy, invoke, and logs
- Works out of the box with sensible defaults

### Deploy to AWS Lambda
`cargo loco lambda deploy`

### Build only (dry run)
`cargo loco lambda deploy --dry-run`

### Invoke the deployed function
```
cargo loco lambda invoke
cargo loco lambda invoke --payload '{"httpMethod": "GET", "path": "/api/users"}'
```

### View CloudWatch logs
```
cargo loco lambda logs
cargo loco lambda logs --follow
```

The handler uses the `lambda_http` crate to adapt HTTP requests from API Gateway or Lambda Function URLs to Axum-compatible requests, which Loco handles natively.

## Dependencies

This feature requires [Cargo Lambda](https://www.cargo-lambda.info/) to be installed:

`cargo install cargo-lambda`

No new Rust dependencies are added to Loco itself—the Lambda runtime dependencies are only added to the user's project during the build process.

## Testing

Tested with a fresh Loco API project:
1. Created new project with `cargo loco new`
2. Configured Lambda settings in `config/development.yaml`
3. Deployed with `cargo loco lambda deploy`
4. Invoked with `cargo loco lambda invoke`
5. Verified logs with `cargo loco lambda logs`

## Related

- Zappa (Python): https://github.com/zappa/Zappa
- Cargo Lambda: https://www.cargo-lambda.info/
- AWS Lambda Rust Runtime: https://github.com/awslabs/aws-lambda-rust-runtime

PR #1806, opened 2026-08-15
Loco 1.1.0

jondot:
Loco 1.1.0.

Moves the template engine to Tera 2, makes configuration files valid YAML,
opens up the storage API, and adds an AWS Lambda deployment target. Full
detail — including every breaking change and its migration — is in
`CHANGELOG.md`.

**Upgrading is a one-line change for most apps:** bump `fluent-templates` from
`0.13` to `0.15`. That crate supplies the `t()` i18n function in generated apps
and pinned Tera 1.

### Shape of this branch

One commit on top of `master`, matching how 1.0.0 landed (#1783). The full
development history is preserved locally on `fix/yaml-safe-config-delimiters`.

### Headlines

- **Tera 2.** Custom filters and functions move to Tera 2 signatures. Loco
  absorbs the rest — it re-registers `get_env` with Tera 1 semantics, and view
  loading yields the same template names, so existing `render()` calls are
  unaffected.
- **Config files are valid YAML at rest** (`<%= ... %>` instead of `{{ ... }}`).
  `{` is a YAML flow-mapping indicator, so any tool reading the file as YAML
  restructured it and broke startup. Legacy `{{ }}` still renders with a
  deprecation warning. (#1727)
- **AWS Lambda deployment generator.** (#1699)
- **Storage gains `list` / `stat` / `exists`** across drivers, strategies and
  the facade. (#1805)
- **Security: `opendal` 0.57 → 0.58.1**, clearing RUSTSEC-2026-0194 and
  RUSTSEC-2026-0195. Reaches only apps enabling a cloud storage feature, all
  off by default. This ships in 1.1.0, not 1.0.x.

### Notable fixes

- **The SQLite queue stopped handing out jobs after `clear()`** — permanently
  and silently. Any app with `queue.dangerously_flush: true` hit it on every
  boot. The advisory-lock table it depended on is gone in favour of `BEGIN
  IMMEDIATE`, which is how SQLite takes the write lock before the SELECT that
  picks a job.
- **Redis: jobs vanished from every operator tool** the moment they completed
  or failed, so `jobs dump --status failed` and `jobs purge` silently did
  nothing.
- **Generated apps pinned a `loco-rs` that could not read their own config**,
  because `cargo xtask bump` had silently failed to maintain the version floor
  since 0.14. A no-match is now an error.
- **A generated migration that could not be registered reported success** —
  fixed at the source in rrgen 0.6, which Loco now requires.
- **Test snapshots broke west of UTC** since 0.14; CI never caught it because
  runners are UTC. (#1802)
- **`config/production.yaml` was generated as a 0-byte file**, and the
  generated `.gitignore` excluded it.

### Verification

`fmt`, `clippy --all-features --all-targets -D warnings`, `test
--all-features`, `loco-gen --all-features`, `xtask docs-syntax`, `snipdoc
check`, `cargo deny`, `hack check --each-feature`, the `loco-new` wizard matrix
against a freshly installed CLI, and both examples.

### Credit

The squash carries `Co-authored-by` trailers for all 24 contributors whose
work is in this release, so GitHub attributes it to them. (1.0.0 shipped
without these — that credit was lost.)


schungx:
Moving to Tera 2 by default should be a major version bump?  Or will this be backwards compatible?

I cannot see how it will be backcompat, because Tera 2 is a complete from-scratch rewrite and a lot of things are completely different from Tera 1.

jondot:
> Moving to Tera 2 by default should be a major version bump? Or will this be backwards compatible?
> 
> I cannot see how it will be backcompat, because Tera 2 is a complete from-scratch rewrite and a lot of things are completely different from Tera 1.

you are correct, and this was a hard dilemma, I wanted to incorporate Tera 2 because it has such great features and additions, but I wouldn't like to go Loco 2.x just yet.

in principle I want to run fast and break things, so hopefully 'breaking changes' will be tamed by AI, I make sure that we have all the material needed for a capable AI agent to comprehend what was changed and be able to fluently migrate.

schungx:
In your migration notes, does it mean the just by merely pinning `fluent-templates` to a lower version I'd be able to force Tera to stay on version 1?

jondot:
Other way around, actually. `fluent-templates` isn't what pulls Tera in — loco depends on `tera 2.1` directly. What that crate gives you is the `t()` function, which your app registers into loco's engine in `src/initializers/view_engine.rs`:

```rust
tera.register_function("t", FluentLoader::new(arc.clone()));
```

With 0.13 you'd end up with tera 1 and tera 2 both in the tree, and that line stops compiling, since 0.13's `FluentLoader` implements tera 1's `Function`. Your templates still go through Tera 2 either way.

So if you need to stay on Tera 1, stay on `loco-rs = "1.0"`. It's not going anywhere.

One thing to watch if you do upgrade: Tera 2 errors on undefined variables where Tera 1 gave you an empty string, and mailer templates use the same instance. So an optional field that's sometimes missing will blow up at send time rather than at compile time. `| default(value="")` on those.

schungx:
> One thing to watch if you do upgrade: Tera 2 errors on undefined variables where Tera 1 gave you an empty string, and mailer templates use the same instance. So an optional field that's sometimes missing will blow up at send time rather than at compile time. | default(value=\"\") on those.

Yes, this is one major thing that would require me to manually go thru all my templates line by line. Not fun. Well... Maybe fun if I use AI... 

Also the fact that there are no more macros, replaced with components. I agree it is much better... But... I have LOTS of macros... Bummer.

PR #1809, opened 2026-08-16
feat(gen): add a user:delete task to generated apps

jondot:
Rebases and lands @floscodes' work from #1708, which had been open since December and predates 1.1.0.

Generated apps ship `user:create` with no way to remove a user again. This adds the symmetric `user:delete` — takes a pid, prints the matched user, confirms before deleting, and `force:true` skips the prompt for scripted use.

Two changes from the original:

- The confirmation read moved to `spawn_blocking`. `stdin().read_line` inline in an `async fn` blocks the runtime thread, and this ships as template code people copy into their own tasks.
- The `detail` string now shows its arguments, matching `user:create`. It previously read `cargo loco run task user:delete` with no hint that `pid` is mandatory.

Left out: the original branch also uncommented the `test_all_combinations` wizard matrix and renamed a join-table fixture. Both are unrelated to the feature — happy to take them separately if wanted.

Verified locally beyond the snapshots: generated an app with auth and ran its own task suite, all three tests pass (`user_create` ×2, `user_delete` ×1).

Closes #1708.

### DIFF
--- a/loco-new/base_template/src/app.rs.t
+++ b/loco-new/base_template/src/app.rs.t
@@ -95,6 +95,7 @@ impl Hooks for App {
         // tasks-inject (do not remove)
         {%- if settings.auth %}
         tasks.register(tasks::user_create::UserCreate);
+        tasks.register(tasks::user_delete::UserDelete);
         {%- endif %} 
     }
 
--- a/loco-new/base_template/src/tasks/mod.rs.t
+++ b/loco-new/base_template/src/tasks/mod.rs.t
@@ -1,3 +1,4 @@
 {%- if settings.auth %}
 pub mod user_create;
+pub mod user_delete;
 {%- endif %}
--- a/loco-new/base_template/src/tasks/user_delete.rs
+++ b/loco-new/base_template/src/tasks/user_delete.rs
@@ -1,0 +1,77 @@
+use crate::models::_entities::users;
+use loco_rs::prelude::*;
+
+pub struct UserDelete;
+#[async_trait]
+impl Task for UserDelete {
+    fn task(&self) -> TaskInfo {
+        TaskInfo {
+            name: "user:delete".to_string(),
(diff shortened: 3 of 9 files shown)


jondot:
Two CI rounds found two real defects, both now fixed:

1. `force` was read with `map(..).unwrap_or(false)` on a `Result`. The wizard tests lint the generated app with clippy pedantic/nursery, and generated apps ship a `.cargo` config making warnings errors — so that's a hard build failure there, not a lint nit. Now `is_ok_and`.

2. Every db-enabled wizard case failed with `no such table: main.groups`. The fixture generates `CreateJoinTableUsersAndGroups`, but `groups` is never created — only `movies` is — so the join table has carried a foreign key to a nonexistent table for a long time. SQLite checks that on write, and nothing in a generated app's suite had ever deleted a row until this task. Join table now points at `movies`.

@floscodes had already fixed #2 on the original branch; I dropped it when porting as 'unrelated to the feature' and was wrong — it's a prerequisite.

Verified locally this time rather than pushing and watching: ran `test_starter_combinations::case_2` end to end, generated app's own suite passes 34/34 including `can_run_user_delete_by_pid`.

jondot:
Windows passed on re-run (2h24m), so that failure was intermittent, not caused by this change. All 15 checks green.

Recording what it was, because it will recur: `SqliteError code 5: "database is locked"` in `user_create` at `boot_test()`. Generated apps point every test at one shared SQLite file, and `boot_test` has no explicit pool close — the connection drops lazily at end of test. On Windows the file lock outlives the test, so the next boot can find it held. All three task tests are `#[serial]`, so this is boot/teardown timing, not concurrency.

Adding a third boot/teardown cycle made an existing race likelier to trip; it didn't create one. Worth a separate look at whether `boot_test` should close its pool deterministically, or whether the generated task tests should use `boot_test_with_create_db` and get their own database.

floscodes:
@jondot

Thank you for picking this up and for the thoughtful improvements. I really appreciate your time and effort!

PR #1797, opened 2026-08-04
fix(website): polish content and interactions

WaterWhisperer:
Fixes #1794

PR #1810, opened 2026-08-17
chore: retire docs-site/ and drop the llms-check gate it served

jondot:
Retires the zola site that `website/` replaced, and removes the CI gate that was propping it up.

### Why now

`docs-site/` stopped being published when the Astro site went live, but stayed in the tree. It carried **9 of the 12 open Dependabot alerts** — including the only critical one — through a lockfile for a site nobody builds, and every website PR had to patch the same page in two trees (#1797 did exactly that).

### The gate that verified nothing

`xtask llms-check` was still pointed at `docs-site/static/*` and `docs-site/content/docs`. So it validated a **42 KB** artifact nobody serves against 64 pages nobody publishes, while the **441 KB** `llms-full.txt` that agents actually fetch from loco.rs had no gate at all.

I started by repointing it at `website/`, then found that doesn't work and shouldn't: `website/public/llms*.txt` are **generated** by `scripts/generate-llms-txt.mjs` as a `prebuild` step, from the same docs tree, and are gitignored. Drift is structurally impossible, so the check has no job — and a repointed version would fail on a fresh checkout, since the `check` job never runs `pnpm prebuild`. Removed rather than moved. The website build still exercises the generator on every run.

### Nothing lost, with one exception

`website/` carries all 64 doc pages plus `how-to/add-a-database` and `how-to/build-a-spa`; blog, casts and authors match.

The exception is **`privacy-policy`**, which has no equivalent in `website/` — it silently stopped being served at the migration, so loco.rs has presumably been 404ing it since. Ported to `website/src/pages/privacy-policy.astro`, text unchanged. Its "contact us" link pointed at `github.com/aaranxu/adidoks`, the zola theme author's repo (template leftover); it now points at Loco's discussions. **Flagging in case you'd rather drop the page entirely than restore it** — that's a call about the live site, not a cleanup.

### Also

`thin-vec 0.2.14 -> 0.2.19` in loco-new's lockfile clears the one high-severity Rust advisory (use-after-free in `IntoIter::drop`, transitive via rhai).

**Alert count: 12 → 2**, both low (rand unsoundness, needs a custom logger to trigger).

### Verified locally

Website builds — 85 pages, `privacy-policy` among them; `prebuild` regenerates both llms files; `docs-syntax` passes; `fmt` clean.

Noted separately, not fixed here: CI's `style` job runs plain `cargo clippy` at the root, which with a root package lints only `loco-rs` — so `xtask` and `loco-gen` have never been pedantic-linted. `xtask` currently has 8 such findings (10 before this PR).


jondot:
CI caught a dependency I missed — my first grep covered `.rs/.yml/.toml/.md` but not `.mjs`. `website/scripts/url-parity.mjs` and `url-parity-blog.mjs` walked `docs-site/content` to learn which URLs loco.rs used to serve, so deleting the tree removed their input.

Those checks are worth keeping — they're what guarantees an old URL still resolves. But the set of legacy URLs is *history*: fixed at the moment of the migration, unable to change again. Deriving it from a live tree meant deleting a page silently deleted the obligation to keep its URL working. It's now frozen in `website/scripts/legacy-urls.json` — 64 doc URLs, 18 blog/casts/authors URLs — and as data the obligation outlives the page.

Verified locally without `docs-site/` present: `url-parity` reports 0 missing against all 64, `url-parity-blog` 0 missing against all 18, build is 85 pages, 39 unit tests pass.

Also noticed but deliberately not touched here: `migrate-docs.mjs` and `migrate-blog-casts.mjs` are one-off migration tools that still point at the old tree, so they're now inert. Their unit tests use fixtures and still pass, so nothing breaks — but they're dead and could go in a follow-up.

PR #1821, opened 2026-08-26
feat(bgworker): perform_all_later batch job enqueueing

### LINKED ISSUE #1803: Bulk worker enqueuing / `perform_all_later`
## Feature Request

Today, if users desire to enqueue many units of work, they must call `perform_later` in a loop, causing a single round-trip write to the queue per job submitted. This could be more efficient by introducing a facility to submit a batch of jobs at once.

A Loco equivalent of Rails's [`ActiveJob::perform_all_later`](https://guides.rubyonrails.org/active_job_basics.html#bulk-enqueuing) seems like a clear addition to the framework. All existing providers should support an atomic batch insert, which would have the affect of all jobs in the group failing if an insert fails.

Something like:
```rust
async fn perform_all_later(ctx: &AppContext, args: Vec<A>) -> Result<String> { 
    perform_all_later_with_priority(ctx, args, vec![]).await
}

async fn perform_all_later_with_priority(ctx: &AppContext, args: Vec<A>, priorities: Vec<Option<i32>>) { ... }
```

Direct queue access via escape hatches like Initializers to add `Extensions` or `SharedStore`, `ExtraDB` etc.

mccormickt:
Adds `perform_all_later` and `perform_all_later_with_priority` functions to batch queue jobs, modeled after Rails's [ActiveJob::perform_all_later](https://guides.rubyonrails.org/active_job_basics.html#bulk-enqueuing). A batch is inserted in a single atomic transaction and therefore can be retried on failure without duplication and returns the `JobID`s of all jobs submitted.

Fixes #1803

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -144,6 +144,11 @@ impl BackgroundWorker<DownloadWorkerArgs> for DownloadWorker {
 let job_id = DownloadWorker::perform_later(&ctx, args).await?;
 // enqueue at a priority (higher runs first), on ANY backend:
 DownloadWorker::perform_later_with_priority(&ctx, args, Some(100)).await?;
+// enqueue many jobs in ONE round trip (atomic on every backend); returns
+// one id per job, in input order:
+let job_ids = DownloadWorker::perform_all_later(&ctx, args_list).await?;
+// same, with a priority per job (None = default):
+DownloadWorker::perform_all_later_with_priority(&ctx, vec![(a, Some(100)), (b, None)]).await?;
 ```
 
 - Backends: Postgres and SQLite ship by default (feature `worker`); Redis needs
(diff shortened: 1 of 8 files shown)


jondot:
Sorry for the wait. The failing `style` check isn't yours: six clippy errors in `cli.rs` and the extractors, files this PR doesn't touch. CI floated on `stable`, 1.98 added two lints, master went red on its own.

Fixed in #1825, which also stops `style` gating `test` and `build` so you get real results instead of SKIPPED. Rebase once it lands and I'll review.

jondot:
Reviewed, and this is in good shape. Atomicity holds on all three backends, the chunk sizes respect both parameter caps, and the default `enqueue_batch` keeps third-party providers compiling. The SQLite `run_at` text formatting is a nice catch — binding a `DateTime<Utc>` there would have hidden every job until the next day.

One request. The change to `perform_later_with_priority` from "log and hand back a fake UUID" to `Err(QueueProviderMissing)` is right, but it's a behavior change unrelated to batching. Could you split it into its own commit? It needs to be visible in the release notes rather than found by someone whose misconfigured app started erroring.

Otherwise ready once #1825 lands and you rebase.

mccormickt:
> Reviewed, and this is in good shape. Atomicity holds on all three backends, the chunk sizes respect both parameter caps, and the default `enqueue_batch` keeps third-party providers compiling. The SQLite `run_at` text formatting is a nice catch — binding a `DateTime<Utc>` there would have hidden every job until the next day.
> 
> One request. The change to `perform_later_with_priority` from "log and hand back a fake UUID" to `Err(QueueProviderMissing)` is right, but it's a behavior change unrelated to batching. Could you split it into its own commit? It needs to be visible in the release notes rather than found by someone whose misconfigured app started erroring.
> 
> Otherwise ready once #1825 lands and you rebase.

Split out the fail-fast on missing queue provider into https://github.com/loco-rs/loco/pull/1821/commits/58395717e79c9b06e88aa027bb777e44eb0b7608

mccormickt:
Friendly ping, @jondot. If you could re-review whenever you get the chance. Thanks!

Issue #1714, opened 2025-12-22
Contribute rhai-loco to this org


schungx:
Would there be any interest to include the [`rhai-loco`](https://github.com/rhaiscript/rhai-loco) repo to this org? Perhaps renamed to `loco-rhai` or merge into `loco-extras`.

It adds arbitrary scripting support for routes as well as script-defiend Tera filters.

kaplanelad:
Why do you think moving it is better? From my perspective, what you did by creating this crate is really cool! You should get the credit, and users can take your work as an example of how to create any initializer they want and integrate it with Loco.

schungx:
No reason in particular.  Just thought it might be better for everything to fall under one umbrella.

I'm fine either way.

jondot:
i think it might be a good idea, might increase visibility for people looking for what they can use that relate to loco

kaplanelad:
I don’t have a strong opinion on this.
@schungx, if you’d like, feel free to transfer this repository and grant us permission on crates.io to publish new versions.

schungx:
> grant us permission on crates.io to publish new versions.

What username should I grant the crates.io permission to?

kaplanelad:
take the name owners from https://crates.io/crates/loco

schungx:
> take the name owners from https://crates.io/crates/loco

Done.

However, I can't create repos on `loco-rs`.  Therefore, I'd try to grant you admin permission on the repo itself, and you can then move it.

The repo link is: https://github.com/rhaiscript/rhai-loco

schungx:
There is this following `axum` extension marker (`ScriptingEngine`) that you may want to consider moving into Loco core instead in order to support more scripting engines in the future...

```rust
/// Type that wraps a scripting engine for use in [`Axum`][axum] handlers.
#[derive(Debug, PartialEq, Eq, Clone)]
pub struct ScriptingEngine<E>(pub E);

...

impl<S, E> FromRequestParts<S> for ScriptingEngine<E> ...
```

Also, right now the scripting engine would take config info under `initializers/scripting`.  You may want to change it to some other config section.

kaplanelad:
@schungx can you tranfer this repo to under loco org?

schungx:
> @schungx can you tranfer this repo to under loco org?

I can't. I need permission to create a repo under the org.

kaplanelad:
if you are enter to the repo settings, scroll down to `Transfer ownership`. you can select our loco org?

[...]

PR #1822, opened 2026-08-31
Add render_component to TeraView.

schungx:
Tera v2 has a new feature: components.  Currently, there is no way to render a Tera v2 component, short of rendering the entire page.

This PR adds the `render_component` method to `TeraView` in addition to the existing `render`.

### DIFF
--- a/src/controller/views/engine.rs
+++ b/src/controller/views/engine.rs
@@ -200,6 +200,45 @@ impl TeraView {
 
         Ok(Self(tera))
     }
+
+    pub fn render_component<S: Serialize>(&self, component: &str, data: S) -> Result<String> {
+        let context = tera::Context::from_serialize(&data)?;
+
+        #[cfg(debug_assertions)]
+        {
+            let mut tera = self
+                .0
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+
+            // Only create a new Tera instance if the view path files have changed
+            if tera.dirty {
+                tracing::warn!(component, "Hot-reloading Tera view engine");
+
+                tera.dirty = false;
+
+                let new_engine =
+                    Self::create_tera_instance(&tera.view_path, tera.post_process.as_ref())?;
+
+                tera.engine = new_engine;
+            }
+
+            Ok(tera
+                .engine
+                .render_component(component, &context, None, false)
+                .map_err(|err| {
+                    Error::string(&format!("Error rendering component {component}: {err:?}"))
+                })?)
+        }
+
+        #[cfg(not(debug_assertions))]
+        Ok(self
+            .0
+            .render_component(component, &context, None, false)
+            .map_err(|err| {
+                Error::string(&format!("Error rendering component {component}: {err:?}"))
+            })?)
+    }
 }
 
 impl ViewRenderer for TeraView {
--- a/src/controller/views/engine.rs
+++ b/src/controller/views/engine.rs
@@ -201,39 +201,44 @@ impl TeraView {
         Ok(Self(tera))
     }
 
-    pub fn render_component<S: Serialize>(&self, component: &str, data: S) -> Result<String> {
-        let context = tera::Context::from_serialize(&data)?;
+    /// Access the Tera view engine instance.
+    #[cfg(not(debug_assertions))]
+    #[inline(always)]
+    fn tera(&self, _key: &str) -> Result<&tera::Tera> {
+        Ok(&self.0)
+    }
+    /// Access the Tera view engine instance.
+    #[cfg(debug_assertions)]
+    #[inline]
+    fn tera(&self, key: &str) -> Result<impl std::ops::Deref<Target = HotReloadingTeraEngine>> {
+        let mut tera = self
+            .0
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
 
-        #[cfg(debug_assertions)]
-        {
-            let mut tera = self
-                .0
-                .lock()
-                .unwrap_or_else(|poisoned| poisoned.into_inner());
+        // Only create a new Tera instance if the view path files have changed
+        if tera.dirty {
+            tracing::warn!(key, "Hot-reloading Tera view engine");
 
-            // Only create a new Tera instance if the view path files have changed
-            if tera.dirty {
-                tracing::warn!(component, "Hot-reloading Tera view engine");
+            tera.dirty = false;
 
-                tera.dirty = false;
+            let new_engine =
+                Self::create_tera_instance(&tera.view_path, tera.post_process.as_ref())?;
 
-                let new_engine =
-                    Self::create_tera_instance(&tera.view_path, tera.post_process.as_ref())?;
-
-                tera.engine = new_engine;
-            }
-
-            Ok(tera
-                .engine
-                .render_component(component, &context, None, false)
-                .map_err(|err| {
-                    Error::string(&format!("Error rendering component {component}: {err:?}"))
-                })?)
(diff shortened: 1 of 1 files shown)


schungx:
Ping.

I have used AI to convert all my Tera v1 templates to Tera v2.  So far the process has been quite smooth, took a few days but much less time than I anticipated.

There are a few gotcha's that I can write up if necessary...

jondot:
Sorry for the wait. Same cause as #1821: the `style` failure isn't yours, CI floated on `stable` and 1.98 added two lints. Fixed in #1825, rebase once it lands.

Two things on the change. The `debug_assertions` branch repeats the hot-reload block from `render` verbatim, worth pulling into a helper both call. And a test would help: one component rendering, one unknown name erroring.

Yes please on the Tera gotchas. We ship an upgrade guide at `docs/extras/upgrades.md` covering filter signatures, undefined variables now erroring, and the `fluent-templates` 0.15 pin. What would help most is what it misses.

Replying on #1714 separately.

schungx:
Gotcha's: Using AI to Convert Tera v1 Templates to Tera v2
==========================================================

Null Handling
-------------

In Tera v1, you can do `<p>{{ text }}</p>` to render a blank tag if `text` does not exist.

In Tera v2, this becomes an error.

The solution is to convert all nullable variable access to `text or ""` which renders fine.

AI is good in doing this modification.


Detecting Types
---------------

In Tera v2, there are now `is array`, `is string` etc.

You no longer have to use `is iterable` and not knowing whether a value is a string or an array etc.

AI should be good in doing this modification.


Patterns Need `pat=`
--------------------

In Tera v2, `is containing` requires `pat=` for the pattern.

Tera v1: `is containing("hello")`

Tera v2: `is containing(pat="hello")`

AI is very good in catching these.


Filters and Maps
----------------

In Tera v2, you can do `[item.key for item in collection if item.value != ""]`.

So Tera v1's `map` and `filter` filters are no longer needed.

AI can rewrite these expressions.


Default Filter
--------------

In Tera v2, `result | default(value=42)` is no longer necessary.

Replace with `result or 42` which is much cleaner and detects if `result` is truthy or falsy.

AI is good in doing this modification.


Missing Filters
---------------

Many built-in filters from Tera v1 no longer load by default.

Some are no longer necessary, e.g. the `json_encode` filter is not needed as maps automatically render as JSON.

The old filters are in the `tera-contrib` crate.  This crate must be added to `Cargo.toml` to use them.


Coalescing Operators
--------------------

Tera v2 has coalescing operators such as `foo?.bar?.baz` and `foo?[42]` etc.

Use them with `or`: `foo?.bar?.baz or "N/A"`.

You no longer need: `{% if not foo or not foo.bar or not foo.bar.baz %}N/A{% else %}{{ foo.bar.baz }}{% endif %}`

AI is quite good in replace these expressions to use the coalescing operators if given clear instructions.


String Comparisons
------------------

Tera v2 now has built-in string comparisons, such as `>=`, `<` etc.

Your own custom filters in Tera v1 are no longer necessary.

If you have them, ask AI to convert them into the standard operators, which it'll do just fine.


Macros -> Components
--------------------

It is surprisingly easy to use AI to convert Tera v1 macros to Tera v2 components.

Simply ask AI to do that, and it'd successfully convert most of them automatically.

Ask the AI to eliminate all `import` statements which are also no longer needed.


Use Components as Values
------------------------

It is OK to use a Tera v2 component as value to a variable, just like Tera v1 macros.

Example: `{% set text = <foo bar={42} title="hello" /> %}`


Component Gotcha's
------------------

### Naming

Previously, macros are scoped to the individual template file.

In Tera v2, components are globally-scoped.

Therefore, you can now use one text name to rendre a component, instead of the file:macro
pair you needed in Tera v1.  The template file name no longe needs to be kept.

Suggested naming is to namespace the components, such as `foo.bar.baz`.

AI can easily do the renaming, or it can be done via a global search-and-replace.


### Varargs

Tera v2 components, different from Tera v1 macros, accept a _fixed_ number of arguments.

Extra, unrecognized arguments will cause an error.

The solution is to put `...rest` at the end of the arguments list, and Tera v2 will put
all unrecognized arguments into that variable (commonly named `rest`) which is a map.

AI is good at doing this.  You can also ask AI to only put `...rest` in components
where it detects additional arguments.


### Argument types

Tera v2 component arguments can be typed, and an argument with a value of the wrong type
will cause an error.

If a type is not specified, it is essentially `any`, which can be anything (except `null`).

If a type is specified, e.g. `option: map`, then it expects an argument of that type.

There is no _union_ typing, so it is either `any` (type omitted) or a fixed type.

Notice that a typed argument is _mandatory_.

AI is mostly OK in putting in these types.


### Default arguments

Tera v2 optional component arguments can be given defaults.

However, the gotcha is that they are also _typed_.

Therefore, an argument `option = true` will type the argument `option` to `bool`, and it
will no longer accept anything other than a `bool`.

There is no way to specify a default for an optional argument without binding its type.


### Rending components

You render a Tera v2 component this way: `<component arg1=... arg2=... />`.

However, a gotcha is that all argument values, other than strings, _MUST BE WRAPPED_ in `{...}`.

For example: `<component arg1="hello" arg2={42} arg3={ {"a":1, "b":2} } arg4={"hello" | caps} />`

Notice that `arg1`'s value does not need to be wrapped in `{...}` because it is a string.

`arg2` is not a string, so it must be wrapped in `{...}`.

`arg3` is a map, and you must put _SPACES_ between the braces, otherwise Tera confuses them to
be `{{ ... }}` interpolations.  So `{ { map } }` for maps.

`arg4` looks like a string, but it is actually an expression with a filter, so it must also be wrapped.

AI is good at doing the right thing, if you specify these rules clearly.  It might be best to start off
by wrapping everything, including strings, e.g. `arg={"hello"}`.  Then as the AI to do a second pass
to simplify wrapped literal strings.

Also, my AI occaionally screws up the `/>` parts, so beware.

schungx:
@jondot   Fine tuned the code, added tests are suggested.

Also, drafted a document on converting from Tera v1 to v2 using AI.

jondot:
`render_component` passes `autoescape: false`, so a component interpolating user data emits raw HTML while `render()` on the same engine escapes it. Tera's own doc example for that call passes `true`. The new tests use plain strings and `42`, so nothing catches it — worth a case asserting `<script>` comes back escaped.

The refactor itself is right; folding the duplicated `#[cfg]` blocks into `tera()` is better than what it replaced.

Separately — that Tera v1→v2 writeup belongs in the docs, not a PR comment. People upgrading to 1.1 are hitting exactly those. Want to add it under `website/src/content/docs/docs/how-to/`?

schungx:
> `render_component` passes `autoescape: false`

Ah, good catch.  I think should add the `autoescape` parameter to `render_component` here and let the user decide.

Or have two versions, `render_component` which escapes, and `render_component_raw` which doesn't.

What do you think?

> Want to add it under website/src/content/docs/docs/how-to/?

OK, I'll put it there.

schungx:
OK, @jondot

`render_component` now defaults to `autoescape = true` since this mirror the behavior of `render` and should be the right thing to do in 99.99% of the case.

Also, added Tera v1 -> v2 migration document under `website/src/content/docs/docs/how-to/migrate-to-tera-v2-with-ai.md`.

jondot:
Thanks, that's the fix. The website build rejected the new page because it had no frontmatter, so I added a title block and switched the headings to `##` like the other how-to pages. I also gave the embedded-assets `TeraView` the same `render_component`. Without it, an app stops compiling when it turns on embedded assets. Merging for 1.2.

PR #1824, opened 2026-09-04
Add tenant query scoping and a full-stack multitenancy example

### LINKED ISSUE #1640: Multi tenant and multi application support
- [ ] A tenant can subscribe to multiple applications
- [ ] An application can be subscribed to by multiple tenants
- [ ] Tenant member management
- [ ] Role and permission management of tenant members
- [ ] Roles can manage specific permissions for multiple currently subscribed applications

shuvroroy:
## Summary

- Add an opt-in `multi-tenancy` feature that enables `with-db` and exposes tenant-scoping helpers through the model module and prelude.
- Add `TenantEntity` to declare an entity's tenant column and tenant ID type.
- Add `TenantQueryExt::in_tenant` to scope Sea-ORM selects, bulk updates, and bulk deletes.
- Add `TenantActiveModelExt::set_tenant` to assign an unset tenant key, accept the same key, and reject overwriting a different key with `ModelError::TenantMismatch` (HTTP 400).
- Document setup, trusted tenant resolution, scoped reads and writes, and limitations.
- Teach the docs syntax checker to handle rustdoc's hidden setup lines while continuing to validate their code.

## Scope

This is the framework-first PR. The full-stack example, its lockfiles, and example-specific CI changes have been removed from the diff. The example will follow in a separate PR.

Tenant context stays explicit at each call site, without request-global or thread-local state. Applications remain responsible for authorizing the tenant, validating related records, and preventing ownership changes through direct or bulk mutations. These helpers do not introduce automatic global filtering or database row-level security.

## Testing

- 13 tenant-focused tests covering query scoping, cross-tenant read/update/delete isolation, tenant assignment, and the HTTP error response.
- Runnable tenant API doctest.
- Four regression tests for hidden rustdoc lines; all 226 checked documentation snippets pass syntax validation.
- Builds with `multi-tenancy` enabled and with `with-db` alone.
- Formatting, Clippy, Rust documentation checks, and website validation.
- All 12 CI checks are passing.

## Related issue

Closes https://github.com/loco-rs/loco/issues/1640


jondot:
Thanks for this, and for marking it draft, since there's a design question worth settling before you go further.

`TenantEntity`, `TenantQueryExt` and `TenantActiveModelExt` are the valuable part, and right now they live in the example, so nobody gets tenant scoping without copying files out of it. If they moved into `loco-rs` behind a feature flag, #1640 gets a real answer and the example shrinks to demonstrating them.

Two other things: the committed `Cargo.lock` and `pnpm-lock.yaml` are a maintenance cost we'd carry forever, and CI would build a full-stack app on every PR. Would you be up for splitting this into the traits first, example second?

shuvroroy:
> Thanks for this, and for marking it draft, since there's a design question worth settling before you go further.
> 
> `TenantEntity`, `TenantQueryExt` and `TenantActiveModelExt` are the valuable part, and right now they live in the example, so nobody gets tenant scoping without copying files out of it. If they moved into `loco-rs` behind a feature flag, #1640 gets a real answer and the example shrinks to demonstrating them.
> 
> Two other things: the committed `Cargo.lock` and `pnpm-lock.yaml` are a maintenance cost we'd carry forever, and CI would build a full-stack app on every PR. Would you be up for splitting this into the traits first, example second?

@jondot Thanks for the feedback. I’ve narrowed this PR to `TenantEntity`, `TenantQueryExt`, and `TenantActiveModelExt` behind an opt-in `multi-tenancy` feature, with focused tests and documentation.

The full-stack example, its lockfiles, and example-specific CI changes have been removed from this PR. I’ll keep this as the framework-first PR and bring the example in a separate follow-up.

jondot:
This is what I was after — traits in `loco-rs`, lockfiles gone, and the `rustdoc_code` expansion in `docs_syntax` is a real fix rather than a workaround for your own examples.

Running it through the full local gate now, I'll come back with results.

jondot:
Sorry for the wait. It went through the full gate on top of everything else going into 1.2 and came back green. Merging.

PR #1827, opened 2026-09-12
feat(storage): add presign_get/presign_put to StoreDriver

robertazzopardi:
## Summary

Adds presigned URL support to the storage layer: `StoreDriver::presign_get`/`presign_put`, wired through `StorageStrategy` (single + replicated) and exposed on the `Storage` facade.

Loco apps that hand clients direct upload/download links (large files, browser-side uploads) currently have to reach into the underlying `opendal` operator themselves, bypassing Loco's storage abstraction. This keeps that use case inside the framework's own API.

## Changes

- `presign_get`/`presign_put` on `StoreDriver`, implemented for the opendal adapter and the null driver
- `Storage` facade methods delegating through the active strategy
- `PresignPutOptions` for setting `Content-Type` on signed uploads
- Docs update for configure-storage how-to

## Test plan

- [x] `cargo build --all-features`
- [x] `cargo test --all-features` (storage module: 72 passed)
- [x] `cargo clippy --all-features --all-targets` (clean, only pre-existing unrelated warnings)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## Unreleased
+
+### Breaking Changes
+
+- **`StoreDriver::presign_put` now takes `PresignPutOptions`** so callers can
+  set `Content-Type` on signed uploads. Pass `PresignPutOptions::default()`
+  for the previous behavior.
+- **`StorageStrategy` gains required `presign_get` and `presign_put`
+  methods.** Custom strategy implementations must implement them. Built-in
+  strategies already do.
+
+### Added
+
+- **`Storage` facade exposes `presign_get` / `presign_put`** (and
+  `*_with_policy` siblings), routing through the selected strategy.
+- **`PresignedRequest::url()`** helper and **`PresignPutOptions`** for signed
+  upload metadata.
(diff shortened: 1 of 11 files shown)


jondot:
The stub-driver tests and the primary-only rule on `ReplicatedStrategy` are exactly right.

One thing to change before I merge: give `StorageStrategy::presign_get`/`presign_put` default bodies that return the same "not supported" error the driver trait uses. As required methods they break every downstream `StorageStrategy`, and both built-in strategies override them anyway — with defaults this is purely additive and the Breaking Changes section goes away.

Smaller ones:

- The first changelog bullet says `presign_put` "now takes `PresignPutOptions`". It's new in this PR, so there's no previous behavior to preserve.
- `PresignPutOptions` and `PresignedRequest` want `#[non_exhaustive]` — we seal public structs (`src/app.rs`, `src/errors.rs`, `src/config/mod.rs`) so adding a field later isn't breaking. The how-to example then needs `..Default::default()`.
- `presign_s3_roundtrip_get_and_put` returns silently when the env vars aren't set, so it always passes. `#[ignore]` makes the skip visible instead.

robertazzopardi:
Added those fixes, definitely safer to not introduce breaking changes.

jondot:
Looks right now. Merging for 1.2.

PR #1828, opened 2026-09-12
feat: optional db-sqlite feature for Postgres-only compile paths

robertazzopardi:
## Summary

Makes `db-sqlite` its own feature (on by default, so nothing changes out of the box) instead of always compiling in with `with-db`. Postgres-only apps can now build with `default-features = false` and skip `libsqlite3-sys`/SQLite entirely, cutting compile time and the C toolchain dependency for deployments that never touch SQLite.

## Changes

- New `db-sqlite` feature gating sea-orm/sqlx SQLite drivers and the SQLite queue worker
- `bgworker`, `config::queue`, `db::connect` adjusted to compile cleanly with the feature off
- Silenced an unreachable-pattern warning that only appears when `db-sqlite` is disabled
- Docs update for feature flags reference

## Test plan

- [x] `cargo build` (default features, includes `db-sqlite`)
- [x] `cargo build --no-default-features --features auth,cli,with-db,cache_inmem,worker` (Postgres-only path)
- [x] `cargo clippy --all-features --all-targets` and with the Postgres-only feature set — both clean
- [x] `cargo test --all-features` — only failures are docker-dependent testcontainers tests (no docker daemon available locally), unrelated to this change


jondot:
This breaks `loco new --db none --bg queue-sqlite`. That combo generates `default-features = false, features = ["cli", "worker"]` alongside a `kind: Sqlite` queue config — with `db-sqlite` off the variant is gone, so the app compiles and then dies at boot on `unknown variant 'Sqlite', expected 'Redis' or 'Postgres'`. Reproduced against your branch from a separate consumer crate.

Fix is in the wizard: `loco-new/src/settings.rs:80`, split `QueueSqlite` out of the shared arm and push `db-sqlite` next to `worker`. An assertion in `loco-new/tests/templates/background.rs` keeps it fixed.

One gotcha worth knowing — `cargo test --no-default-features` can't catch this here. `[dev-dependencies] loco-rs = { path = "." }` re-enables defaults in any test build of the crate, so only an external consumer ever sees it.

Everything else looks good and I want this in 1.2.

robertazzopardi:
Ah yeah, I should have done more qa with the features on, should be fixed now

jondot:
That fixes it, and the test covers the exact combination. Merging for 1.2.

PR #1829, opened 2026-09-14
docs: modernize "the tour" applying new code style

Data5tream:
Naming conventions and generated code changed after "The Tour" tutorial was written. This PR applies the new conventions and auto-gen code to the tutorial.

- Scaffold automatically adds auth to generated endpoints, use the `--no-auth` flag for the `posts` endpoints.
- Generated code now uses `create` instead of `add` for the creation function and the order of endpoints is different.
- Initialize the data struct with data from params, like in generated endpoints.


jondot:
Checked all of it against the current generator — `--no-auth`, `create`, the `/api/posts` and `/{id}` prefixes, `StatusCode::CREATED`, `CreatePost`. All correct. Merging on green.

PR #1831, opened 2026-09-17
fix(bgworker/redis): survive a slow or dropped worker connection

mikosco4real:
A worker is the only thing draining its queue, so any path that ends the task silently loses every job behind it — while the process stays up and answers health checks. Three such paths exist today, and the `redis` 1.x defaults made two of them reachable.

We hit all three in production this week; details below are from that deployment.

## 1. The connect timeout is inherited, and it changed

| | `redis` 0.x | `redis` 1.x |
|---|---|---|
| `AsyncConnectionConfig` | `#[derive(Default)]` | hand-written `impl Default` |
| `connection_timeout` | **`None`** — waits as long as needed | **`Some(1s)`** |
| `response_timeout` | **`None`** | **`Some(500ms)`** |

`run` calls the bare `get_multiplexed_async_connection()`, which takes `AsyncConnectionConfig::new()` — so it inherits whichever default ships. One second has to cover TCP, AUTH **and** the multiplexer handshake, since `connection_timeout` wraps the whole inner connect rather than just the socket.

**Observed:** on a 256MB shared-CPU instance against a managed Redis, the worker failed at *every* boot with `timed out`, one second after migrations began — while a raw TCP connect from the same host was instant, and the identical call succeeded from the readiness probe moments later. The app looked perfectly healthy and silently processed nothing.

This PR sets both timeouts explicitly. Generous rather than absent: an unreachable Redis still fails, just not on a budget tuned for a local server.

## 2. A failed connect ended the worker

```rust
Err(err) => {
    error!(err = err.to_string(), "Failed to create worker connection");
    return;   // the task exits; nothing ever drains the queue
}
```

One unlucky moment at startup disabled background processing for the life of the process. Now: five attempts with exponential backoff (250ms doubling to a 5s cap), cancellation-aware, with a distinct message when it genuinely gives up.

## 3. A dropped connection was never replaced

The connection is created once outside the loop, and `dequeue_with_conn` errors were logged and swallowed — so a socket that dies mid-flight produces the same error on every poll, forever, while the worker consumes nothing.

**Observed** in the same deployment: `cannot fetch from queue err="broken pipe"` twice a second for 21 hours. Redis was healthy the whole time; `AUTH`/`PING` from that machine succeeded on demand. The connection is now re-established through the same helper.

## Tests

Two, neither needing a container:

- `connect_worker_honours_cancellation` — a cancelled worker makes no attempt at all
- `connect_worker_retries_before_giving_up` — an unreachable server is retried rather than abandoned on the first failure

## Checks

- `cargo +nightly fmt --all` clean (per CONTRIBUTING)
- `cargo clippy --features worker_redis --all-targets -- -D warnings -W clippy::pedantic -W clippy::nursery -W rust-2018-idioms` reports nothing in `src/bgworker/redis.rs`. It does report 7 pre-existing items in `src/controller/views/engine.rs`, untouched here.

## Notes for reviewers

- The constants are hardcoded rather than plumbed through `RedisQueueConfig`. Happy to make them configurable if you'd prefer — I kept the surface minimal, but a `connect_timeout` alongside the Postgres queue's existing one would be a natural home.
- The same inherited-default question applies to `get_connection()` (used by `ping`, `clear`, and the admin paths). Those are short-lived request-scoped calls where a 1s budget is more defensible, so I left them alone — but happy to align them if you'd rather they were explicit too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jondot:
Thanks for this, and for the writeup. It made the timeout change easy to check. I added one change in the release: after five failed reconnects the worker still exited, and a managed Redis failover can take longer than those 8 seconds. Now it retries until shutdown, with backoff capped at 5s. Merging for 1.2.

